PR #7, opened 2016-09-28
API Proposal: Updated APIs for channel member refactor

jwilander:
Proposal for API changes in mattermost/platform#4106

### DIFF
--- a/source/channels.yaml
+++ b/source/channels.yaml
@@ -191,7 +191,7 @@
           type: string
       responses:
         '200':
-          description: The channel stats
+          description: The channel member
           schema:
             $ref: '#/definitions/ChannelMember'
         '500':
--- a/source/channels.yaml
+++ b/source/channels.yaml
@@ -167,7 +167,7 @@
         '500':
           description: Could not retrieve the channel stats, the user is not a member of the channel or the channel does not belong to the team
 
-  '/teams/{team_id}/channels/{channel_id}/member/{user_id}':
+  '/teams/{team_id}/channels/{channel_id}/members/{user_id}':
     get:
       tags:
         - channels
--- a/source/channels.yaml
+++ b/source/channels.yaml
@@ -142,3 +142,58 @@
         '500':
           description: Could not retrieve the channel, the user is not a member of the channel or the channel does not belong to the team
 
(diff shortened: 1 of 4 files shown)


source/channels.yaml
```diff
@@ -142,3 +142,58 @@
...
+          required: true
+          type: string
+        - name: user_id
+          in: path
+          description: The user ID
+          required: true
+          type: string
+      responses:
+        '200':
+          description: The channel stats
```
enahum: should this be `channel member` instead of channel stats?

source/definitions.yaml
```diff
@@ -105,6 +105,14 @@ definitions:
       creator_id:
         type: string
 
+  ChannelStats:
+    type: object
+    properties:
+      id:
```
enahum: I prefer to be more consistent here maybe as with ChannelMember where `id` is actually `channel_id`?

source/users.yaml
```diff
@@ -114,8 +115,101 @@
...
+      description: Get a map of sanitized user objects that are on the team but not in the channel.
+      parameters:
+        - name: team_id
+          in: path
+          description: The team ID
+          required: true
+          type: string
+        - name: channel_id
+          in: path
+          description: The channel ID
```
enahum: `The channel ID the users are not part of` ? to be more explicit

source/definitions.yaml
jwilander: That was from the legacy stuff, but yes that's a good idea. I'll update

source/users.yaml
jwilander: Agreed, will update

source/channels.yaml
jwilander: Oops, that's right

jwilander:
@enahum Changes made, take a look

jwilander:
I made some updates to the new APIs:
* Changed `/users/profiles_by_ids` to `/users/ids` to be more consistent with REST and to be consistent with future similar APIs.
* Changed `/teams/{team_id}/channels/{channel_id}/member/{user_id}` to '/teams/{team_id}/channels/{channel_id}/members/{user_id}' to keep consistent resource name for `members` APIs

enahum:
+1

PR #9, opened 2016-10-17
Add API for autocompleting users

jwilander:
API proposal for mattermost/platform#4239

### DIFF
--- a/source/users.yaml
+++ b/source/users.yaml
@@ -412,3 +412,33 @@
             application/json:
               email: someone@example.com
 
+  '/teams/{team_id}/users/autocomplete':
+    get:
+      tags:
+        - users
+      summary: Autocomplete users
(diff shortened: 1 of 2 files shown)


source/users.yaml
```diff
@@ -412,3 +412,33 @@
...
+        - name: team_id
+          in: path
+          description: The team ID the users are in
+          required: true
+          type: string
+        - name: term
+          in: query
+          description: The term to match against username, full name or nickname
+          required: true
+          type: integer
```
enahum: I think you meant **string**?

source/users.yaml
```diff
@@ -412,3 +412,33 @@
...
+        - name: channel_id
+          in: query
+          description: The channel ID to autocomplete for
+          required: false
+          type: integer
```
enahum: same here

source/users.yaml
jwilander: Woops..

jwilander:
@enahum fixed

jwilander:
@enahum please re-review at your leisure

PR #12, opened 2016-10-25
PLT-4430 improve channel switching

enahum:
This changes the API's to return only the channels for `getChannels` and `getMoreChannels` and adds a new API `/channels/unread` that returns only the unread messages and mentions count for the channels if there are any


jwilander:
I need to update this to match changes in the platform PR

jwilander:
Closing to resubmit

PR #14, opened 2016-11-09
PLT-1378 Added documentation for reaction routes

hmhealey:


enahum:
@hmhealey needs rebase

hmhealey:
Done

source/reactions.yaml
```diff
@@ -0,0 +1,111 @@
+  '/teams/{team_id}/channels/{channel_id}/posts/{post_id}/save_reaction':
```
jwilander: If reactions are a new object/resource we should treat them like the other resources so maybe make the route this:
`/teams/{team_id}/channels/{channel_id}/posts/{post_id}/reactions/create`

With similar changes to the other ones

source/reactions.yaml
```diff
@@ -0,0 +1,111 @@
+  '/teams/{team_id}/channels/{channel_id}/posts/{post_id}/save_reaction':
+    post:
+      tags:
+        - posts
+      summary: React to a post
+      description: Saves a reaction made by the current user to the given post. Post reactions require Mattermost 3.6 or higher.
```
jwilander: Don't need to specify the Mattermost version, this will be taken care of shortly when I add separate pages for different API versions

source/reactions.yaml
hmhealey: That's a good idea. Will do

hmhealey: I'm going to make this route `/reactions/save` just because duplicate calls are considered as updates  (even though there's nothing to update with reactions) instead of errors

PR #30, opened 2016-11-17
PLT-4167 Team Sidebar

enahum:


source/teams.yaml
```diff
@@ -42,6 +42,20 @@
           schema:
             $ref: '#/definitions/TeamMap'
 
+  /teams/my_members:
```
jwilander: I think the channel equivalent is `/teams/{team_id}/channels/members`, maybe we copy that and do `/teams/members` for this?

source/teams.yaml
```diff
@@ -42,6 +42,20 @@
...
+    get:
+      tags:
+        - teams
+      summary: Get all the teams the user is a member of
+      description: Get all the teams the user is member along with the unread count for messages and mentions for each team
+      responses:
+        '200':
+          description: Teams Found
```
jwilander: Super minor but we `Found` should probably be `found` :P

enahum:
@jwilander added the second endpoint as discussed

PR #32, opened 2016-11-21
Add documentation for PLT-3762 - Add API call to get a user by their …

cpanato:
…username

Platform PR: https://github.com/mattermost/platform/pull/4611

### DIFF
--- a/source/users.yaml
+++ b/source/users.yaml
@@ -188,6 +188,24 @@
             additionalProperties:
               $ref: "#/definitions/User"
 
+  '/users/name/{username}':
+    get:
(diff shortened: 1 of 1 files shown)


source/users.yaml
```diff
@@ -188,6 +188,24 @@
             additionalProperties:
               $ref: "#/definitions/User"
 
+  '/users/name/{username}':
+    get:
+      tags:
+        - users
+      summary: Get user object by username
+      description: Get a map of sanitized user object by a username.
```
enahum: does this return a map or an object?

PR #33, opened 2016-11-21
Add channel delete

MartinDelille:
I'm not really sure if I documented well the responses description/type

### DIFF
--- a/source/channels.yaml
+++ b/source/channels.yaml
@@ -215,7 +215,11 @@
       responses:
         '200':
           description: The channel ID
-          type: string
+          schema:
+            type: object
+            properties:
+              id:
+                type: string
 
   '/teams/{team_id}/channels/{channel_id}/add':
     post:
--- a/source/channels.yaml
+++ b/source/channels.yaml
@@ -200,7 +200,7 @@
       tags:
         - channels
       summary: Delete a channel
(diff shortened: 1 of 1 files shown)


source/channels.yaml
```diff
@@ -195,6 +195,28 @@
         '500':
           description: Could not retrieve the channel stats, the user is not a member of the channel or the channel does not belong to the team
 
+  '/teams/{team_id}/channels/{channel_id}/delete':
+    post:
+      tags:
+        - channels
+      summary: Delete a channel
+      description: Add a channel from a team.
```
jwilander: I think you missed updating the description

source/channels.yaml
```diff
@@ -195,6 +195,28 @@
...
+          type: string
+        - name: channel_id
+          in: path
+          description: The channel ID
+          required: true
+          type: string
+      responses:
+        '200':
+          description: The channel ID
+          type: string
```
jwilander: I think the response is actually an object like so:
```
{
    "id": "idgoeshere"
}
```

MartinDelille: Ok how can I document that?

MartinDelille:
Done

jwilander:
Much thanks!

PR #51, opened 2016-12-02
Add API call to get a channel by its name

cpanato:
part of this ticket https://github.com/mattermost/platform/issues/4515

### DIFF
--- a/source/channels.yaml
+++ b/source/channels.yaml
@@ -97,6 +97,31 @@
           '500':
             description: Could not retrieve the user channels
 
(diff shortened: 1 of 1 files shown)


source/channels.yaml
```diff
@@ -97,6 +97,31 @@
           '500':
             description: Could not retrieve the user channels
 
+  '/teams/{team_id}/channels/name/{channel_name:[A-Za-z0-9_-]+}':
+      get:
+        tags:
+          - channels
+        summary: Get channel by channel name
```
jwilander: Super minor, but what do you think about `Get channel by name` instead?

cpanato: done

PR #53, opened 2016-12-05
PLT-4920 Copy post to another channel

enahum:


source/posts.yaml
```diff
@@ -311,3 +311,41 @@
           schema:
             $ref: "#/definitions/PostList"
 
+  '/teams/{team_id}/channels/{channel_id}/posts/{post_id}/copy':
```
jwilander: Do we really need another API for this? Why not just create a new post in the other channel?

enahum:
totally makes sense, nice one...

I'mm closing this PR and I'll change the other one :)

PR #57, opened 2016-12-19
PLT-5012 Add view channel API

jwilander:


source/channels.yaml
```diff
@@ -73,6 +73,43 @@
...
+                description: The channel ID of the previous channel, used when switching channels. Providing this ID will cause push notifications to clear on the channel being switched to.
+              time:
+                type: string
+                description: The Unix time, in milliseconds, that the channel was viewed at. Provide a non-zero value for this parameter to use a manual time. If this parameter is zero or not included, the server will use the current time.
+
+      responses:
+        '200':
+          description: Channel viewed successfully.
+          schema:
+            $ref: '#/definitions/StatusOK'
```
enahum: Wondering if this should return the timestamp set to the lastViewAt in the channel

jwilander: What do you use the time for on the RN app? I see it getting stored, but where is it used?

enahum: I was planing to use that for the `new messages` indicator, but we can leave it for now and come back to this later if needed

PR #58, opened 2016-12-22
API endpoint to get channel members by IDs.

grundleborg:
The purpose of this is to make it possible to find out the roles of the members of a channel.

This will partially enable PLT-5049.

This is designed to be analogous to the Team Members By Ids API endpoint.

### DIFF
--- a/source/channels.yaml
+++ b/source/channels.yaml
@@ -343,6 +343,39 @@
         '500':
           description: Could not retrieve the channel member, the user is not a member of the channel or the channel does not belong to the team
 
+  '/teams/{team_id}/channels/{channel_id}/members/ids':
+      post:
+        tags:
+          - channels
+        summary: Get channel members by IDs
+        description: Get a map of channel member objects for the specified channel and user IDs.
+        parameters:
+          - name: team_id
+            in: path
+            description: Team ID of the member to return
+            required: true
+            type: string
+          - name: channel_id
+            in: path
+            description: The channel ID
+            required: true
+            type: string
+          - in: body
+            name: body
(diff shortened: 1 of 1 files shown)


source/channels.yaml
```diff
@@ -343,6 +343,39 @@
...
+  '/teams/{team_id}/channels/{channel_id}/members/ids':
+      post:
+        tags:
+          - channels
+        summary: Get channel members by IDs
+        description: Get a map of channel member objects for the specified channel and user IDs.
+        parameters:
+          - name: team_id
+            in: path
+            description: Team ID of the member to return
```
jwilander: Maybe just change this to `Team ID for the corresponding channel` or something similar?

source/channels.yaml
```diff
@@ -343,6 +343,39 @@
...
+            required: true
+            schema:
+              type: array
+              items:
+                type: string
+        responses:
+          '200':
+            description: Members retreived successfully
+            schema:
+              type: object
```
enahum: does this return a map or an array?

jwilander: Map is consistent with the similar endpoint for team members

grundleborg: Good question! It looks like the equivalent method for TeamMembers returns an Array in the code, but claims to return a Map in the API reference.

I'll change this one to an array too, unless you think it makes more sense to change things to a map here?

jwilander: Oh, if it's array in code then yeah let's switch it over for this and can you also update the documentation for the team members one to be accurate?

PR #60, opened 2016-12-22
Add docs for list_team_commands

ZJvandeWeg:


source/teams.yaml
```diff
@@ -284,3 +284,20 @@
...
 
+  '/teams/{team_id}/commands/list_team_commands':
+    get:
+      tags:
+        - teams
+      summary: Get all slash commands for this team
+      parameters:
+        - name: team_id
+          in: path
+          description: Team ID of the team you want stats for
```
jwilander: I think this should be `Team ID of the team you want commands for`

ZJvandeWeg: @jwilander Fixed!

jwilander:
That was quick!

PR #65, opened 2016-12-30
Fix websocket endpoint path

trarbr:
In the implementation, the endpoint for the websocket is `/api/v3/users/websocket` (see [source](https://github.com/mattermost/platform/blob/master/api/websocket.go#L20)). This fixes the docs to match the current implementation.


jwilander:
Thanks for the correction!

PR #66, opened 2017-01-03
PLT-5050 New API to update channel member roles.

grundleborg:
New API endpoint to allow the update of ChannelMember roles. This will enable people to be promoted/demoted to Channel Admin role in the WebApp as per https://mattermost.atlassian.net/browse/PLT-5050

### DIFF
--- a/source/channels.yaml
+++ b/source/channels.yaml
@@ -410,6 +410,44 @@
               additionalProperties:
                 $ref: "#/definitions/ChannelMember"
 
+  '/teams/{team_id}/channels/{channel_id}/update_member_roles':
+      post:
+        tags:
+          - channels
+        summary: Update the roles of a channel member
+        description: Update the roles of the provided user in the specified channel.
+        parameters:
+          - name: team_id
+            in: path
+            description: Team ID for the corresponding channel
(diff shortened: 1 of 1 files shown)


source/channels.yaml
```diff
@@ -410,6 +410,44 @@
               additionalProperties:
                 $ref: "#/definitions/ChannelMember"
 
+  '/teams/{team_id}/channels/{channel_id}/update_member_roles':
```
jwilander: To get a channel member, the current API is `/teams/{team_id}/channels/{channel_id}/members/{user_id}` so maybe go with `/teams/{team_id}/channels/{channel_id}/members/{user_id}/update_roles`? Also matches `/users/{user_id}/update_roles` as well

grundleborg: I went for consistency with /teams/{team_id}/update_member_roles instead, but happy to change to be consistent with those other ones instead if you prefer?

jwilander: Hmmm, ok yeah let's be consistent with the `/teams/{team_id}/update_member_roles` since that's the closest endpoint to this one

PR #81, opened 2017-02-02
Document DELETE /users/{user_id} endpoint for APIv4

ruzette:


source/v4/users.yaml
```diff
@@ -122,4 +122,29 @@
...
         '401':
           $ref: '#/responses/Unauthorized'
+    delete:
+      tags:
+        - users
+      summary: Deactivate a user account.
+      description: |
+        Deactivates the user by archiving its user object.
+        ##### Permissions
+        Must be logged in as the user being updated or have the `edit_other_users` permission.
```
jwilander: Really minor but can you change
```
Must be logged in as the user being updated
```
to
```
Must be logged in as the user being deactivated
```

source/v4/users.yaml
```diff
@@ -122,4 +122,29 @@
...
+          type: string
+      responses:
+        '200':
+          description: User deactivation successful
+          schema:
+            $ref: '#/definitions/StatusOK'
+        '400':
+          $ref: '#/responses/BadRequest'
+        '401':
+          $ref: '#/responses/Unauthorized'
```
jwilander: Can you also add
```
'403':
      $ref: '#/responses/Forbidden'
```
since I think that is a possible error code for this endpoint?

ruzette: Updated! I didn't notice that I had put a forbidden request on the unit test. Thanks for catching that!

jwilander:
Thanks for the updates!

PR #87, opened 2017-02-05
Document GET /users/username/{username} endpoint for APIv4

ruzette:


source/v4/users.yaml
```diff
@@ -233,3 +233,29 @@
...
+        - users
+      summary: Get a user by username
+      description: |
+        Get a user object by providing a username. Sensitive information will be sanitized out.
+        ##### Permissions
+        Requires an active session but no other permissions.
+      parameters:
+        - name: username
+          in: path
+          description: User Name
```
jwilander: Like in the other PR, can we make this `Username`? :)

PR #89, opened 2017-02-10
Document some post endpoints for APIv4

jwilander:


source/v4/posts.yaml
```diff
@@ -0,0 +1,136 @@
...
+      tags:
+        - posts
+      summary: Create a post
+      description: |
+        Create a new post in a channel. To create the post as a comment on another post, provide `root_id`.
+        ##### Permissions
+        Must have `create_post` permission for the channel the post is being created in.
+      parameters:
+        - in: body
+          name: body
```
enahum: is the name of the parameter body? shouldn't be the object as the whole body?

jwilander: Whoops, yeah that should be `post`

PR #98, opened 2017-02-17
Document user session endpoints for APIv4

ruzette:


source/v4/users.yaml
```diff
@@ -485,3 +485,70 @@
           $ref: '#/responses/Unauthorized'
         '404':
           $ref: '#/responses/NotFound'
+
+  /users/{user_id}/sessions:
+    get:
+      tags:
+        - users
+      summary: Get a list of sessions by User GUID.
```
jwilander: Can we simplify the summary to just be `Get a user's sessions`? It'll be showing up in a sidebar so it's better if we can keep the summaries short :)

source/v4/users.yaml
```diff
@@ -485,3 +485,70 @@
...
+        '401':
+          $ref: '#/responses/Unauthorized'
+        '403':
+          $ref: '#/responses/Forbidden'
+ 
+  /users/{user_id}/sessions/revoke:
+    post:
+      tags:
+        - users
+      summary: Get a list of sessions by User GUID.
```
jwilander: I think this should be `Revoke a session` right?

source/v4/users.yaml
```diff
@@ -485,3 +485,70 @@
...
+      description: |
+        Get a list of sessions by providing the user GUID. Sensitive information will be sanitized out.
```
jwilander: I think you forgot to update this too

source/v4/users.yaml
```diff
@@ -485,3 +485,70 @@
...
+            type: object
+            required:
+              - session_id
+            properties:
+              session_id:
+                description: The session GUID to be revoked.
+                type: string
+      responses:
+        '200':
+          description: User session retrieval successful
```
jwilander: Here as well

ruzette:
Oops. apologies about that. I thought I pushed that commit prior to the PR. @jwilander ready for review!

PR #104, opened 2017-02-20
Added docs for delete and search posts for apiv4

ruzette:


source/v4/posts.yaml
```diff
@@ -70,6 +70,31 @@
           $ref: '#/responses/Unauthorized'
         '403':
           $ref: '#/responses/Forbidden'
+    delete:
+      tags:
+        - posts
+      summary: Deletes a post
```
jwilander: Minor but can we make this `Delete a post`?

source/v4/posts.yaml
```diff
@@ -70,6 +70,31 @@
...
+      description: |
+        Deletes a single post.
+        ##### Permissions
+        Must be logged in as the user or have `manage_system` permission.
```
jwilander: I think it's actually "logged in as the user who created the post or have the `delete_others_posts` permission." right?

source/v4/posts.yaml
```diff
@@ -70,6 +70,31 @@
...
+      parameters:
+        - name: post_id
+          in: path
+          description: ID of the post to get
```
jwilander: Should be "ID of the post to delete"

source/v4/posts.yaml
```diff
@@ -70,6 +70,31 @@
           $ref: '#/responses/Unauthorized'
         '403':
           $ref: '#/responses/Forbidden'
+    delete:
+      tags:
+        - posts
+      summary: Deletes a post
+      description: |
+        Deletes a single post.
```
jwilander: Can you update this to be "Soft deletes a post, by marking the post as deleted in the database. Soft deleted posts will not be returned in post queries."? Later we'll add endpoints for permanently deleting posts

jwilander:
@ruzette can you give this a rebase when you get a chance?

ruzette:
@jwilander rebased and updated the descriptions per feedback. Ready for re-review. :)

PR #108, opened 2017-02-23
add reference doc for Check if team exists endpoint for APIv4

cpanato:
API PR: https://github.com/mattermost/platform/pull/5517

### DIFF
--- a/source/v4/teams.yaml
+++ b/source/v4/teams.yaml
@@ -98,6 +98,31 @@
         '404':
           $ref: '#/responses/NotFound'
 
+  '/teams/name/{name}/exists':
+    get:
+      tags:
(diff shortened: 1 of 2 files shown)


source/v4/teams.yaml
```diff
@@ -98,6 +98,31 @@
         '404':
           $ref: '#/responses/NotFound'
 
+  '/teams/name/{name}/exists':
+    get:
+      tags:
+        - teams
+      summary: Check if Team exists
```
jwilander: Can we drop the capitalization on `Team` so it's `Check if team exists`?

cpanato: done

PR #111, opened 2017-02-24
Put user mfa api4

andreistanciu24:
docs for "PUT /users/{user_id}/mfa"

### DIFF
--- a/source/v4/users.yaml
+++ b/source/v4/users.yaml
@@ -552,3 +552,47 @@
           $ref: '#/responses/Unauthorized'
         '403':
           $ref: '#/responses/Forbidden'
+
+/users/{user_id}/mfa:
+    put:
+      tags:
+        - users
+      summary: Update a user's mfa
+      description: |
+        Activates multi-factor authenticates for the current user if activate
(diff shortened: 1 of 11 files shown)


source/v4/users.yaml
```diff
@@ -581,3 +581,48 @@
...
         '403':
           $ref: '#/responses/Forbidden'
+
+  /users/{user_id}/mfa:
+    put:
+      tags:
+        - users
+      summary: Update a user's mfa
+      description: |
+        Activates multi-factor authenticates for the current user if activate
```
jwilander: Can this go from `Activates multi-factor authenticates for the current user` to `Activates multi-factor authentication for the user`?

source/v4/users.yaml
```diff
@@ -581,3 +581,48 @@
...
+            properties:
+              activate:
+                description: Activation status
+                type: string
+              token:
+                description: Validation token
+                type: string
+      responses:
+        '200':
+          description: User mfa update successful
```
jwilander: Can we update this to `User MFA update successful`?

andreistanciu24:
done

PR #115, opened 2017-02-27
Get unread totals for all the teams a user is on.

andreistanciu24:
GET /users/{user_id}/teams/unread

### DIFF
--- a/source/v4/teams.yaml
+++ b/source/v4/teams.yaml
@@ -335,7 +335,7 @@
     get:
       tags:
         - teams
-      summary: Get the count for unread messages and mentions in the teams the user is a member of
+      summary: Get team unreads for a user
       description: |
         Get the count for unread messages and mentions in the teams the user is a member of.
         ##### Permissions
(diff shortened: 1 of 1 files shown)


source/v4/teams.yaml
```diff
@@ -271,3 +271,37 @@
           $ref: '#/responses/Forbidden'
         '404':
           $ref: '#/responses/NotFound'
+
+  /users/{user_id}/teams/unread
+    get:
+      tags:
+        - teams
+      summary: Get the count for unread messages and mentions in the teams the user is a member of
```
jwilander: Can we shorten this to `Get team unreads for a user`? It's better to keep the summaries short since they'll appear in a sidebar

source/v4/teams.yaml
```diff
@@ -271,3 +271,37 @@
...
+          required: true
+          type: string
+        - name: exclude_team
+          in: query
+          description: Optional team id to be excluded from the results
+          required: true
+          type: string
+      responses:
+        '200':
+          description: Teams found
```
jwilander: Can we change this to `Team unreads retrieval successful`?

andreistanciu24:
done

PR #116, opened 2017-02-27
ApiV4: GET /teams

andreistanciu24:


source/v4/teams.yaml
```diff
@@ -40,6 +40,36 @@
         '403':
           $ref: '#/responses/Forbidden'
 
+    '/teams':
+    get:
+      tags:
+        - teams
+      summary: Get a list of all teams.
```
jwilander: Can you change this to `Get teams`?

source/v4/teams.yaml
```diff
@@ -40,6 +40,36 @@
...
+      description: |
+        Logged in, only shows open teams. Shows all teams with "manage_system" permission. The result is based on query string parameters - page and per_page.
```
jwilander: Can you change this to `For regular users only returns open teams. Users with the "manage_system" permission will return teams regardless of type. The result is based on query string parameters - page and per_page.`

source/v4/teams.yaml
```diff
@@ -40,6 +40,36 @@
...
+        - name: per_page
+          in: query
+          description: The number of teams per page.
+          default: "60"
+          type: string
+      responses:
+        '200':
+          description: Team list retrieval successful
+          schema:
+            $ref: '#/definitions/Team'
```
jwilander: This needs to be an array, see other endpoints for examples

andreistanciu24:
done.

PR #126, opened 2017-03-03
ApiV4: POST /channels/{channel_id}/members/ids

andreistanciu24:


source/v4/channels.yaml
```diff
@@ -199,6 +199,42 @@
...
+        '200':
+          description: Channel member list retrieval successful
+          schema:
+            type: array
+            items:
+              $ref: '#/definitions/ChannelMember'
+        '400':
+          $ref: '#/responses/BadRequest'
+        '401':
+          $ref: '#/responses/Unauthorized'
```
jwilander: Can you add for the forbidden status code?

andreistanciu24: done.

PR #128, opened 2017-03-03
ApiV4: GET /channels/{channel_id}/unread

andreistanciu24:


source/v4/channels.yaml
```diff
@@ -350,6 +350,35 @@
         '403':
           $ref: '#/responses/Forbidden'
 
+ '/channels/{channel_id}/unread':
+    get:
+      tags:
+        - channels
+      summary: Get the count for unread messages
```
jwilander: Can we make this `Get channel unreads`?

source/v4/channels.yaml
```diff
@@ -350,6 +350,35 @@
...
+      description: |
+        Get unread totals for a channel.
+        ##### Permissions
+        Must be logged in as user or the `manage_system` permission.
```
jwilander: Permission should check `read_channel`.

PR #131, opened 2017-03-06
Document SAML endpoints for APIv4

jwilander:


jwilander:
@enahum can you give this a review as well when you get a chance? Maybe just make sure I got the terminology correct and it all makes sense

source/v4/saml.yaml
```diff
@@ -0,0 +1,182 @@
...
+          $ref: '#/responses/Forbidden'
+        '501':
+          $ref: '#/responses/NotImplemented'
+
+    delete:
+      tags:
+        - SAML
+      summary: Remove IDP certificate
+      description: |
+        Delete the current IDP certificate being used with your SAML configuration. This will also disable SAML on your system as this certificate is required for SAML.
```
enahum: This shouldn't disable SAML as there's the option to upload the file again, im i right?

source/v4/saml.yaml
```diff
@@ -0,0 +1,182 @@
...
+          $ref: '#/responses/Forbidden'
+        '501':
+          $ref: '#/responses/NotImplemented'
+
+    delete:
+      tags:
+        - SAML
+      summary: Remove public certificate
+      description: |
+        Delete the current public certificate being used with your SAML configuration. This will also disable SAML on your system as this certificate is required for SAML.
```
enahum: Definitely not true as this certificate is to be use to verify the if the SAML response is a valid response, you can have your SAML configured without the need the verification step

source/v4/saml.yaml
```diff
@@ -0,0 +1,182 @@
...
+          $ref: '#/responses/Forbidden'
+        '501':
+          $ref: '#/responses/NotImplemented'
+
+    delete:
+      tags:
+        - SAML
+      summary: Remove private key
+      description: |
+        Delete the current private key being used with your SAML configuration. This will also disable SAML on your system as this key is required for SAML.
```
enahum: as before this one is used to decrypt the SAML assertions so you can actually have SAML configured without this, the only thing is that if you enable encryption you will need the public certificate too

source/v4/saml.yaml
jwilander: I think you're thinking of it from the UI perspective, which we shouldn't do with the API. If this file is removed I can't save the config file unless SAML disabled because it fails the IsValid check. The UI can just re-enable SAML in the config when they upload a new certificate

source/v4/saml.yaml
jwilander: Ok my mistake, I'll remove the disabling for this

jwilander:
@enahum changes made here and on the platform PR

PR #141, opened 2017-03-11
Document GET /cluster/status endpoint for APIv4

jwilander:


jwilander:
Yes, that's a good idea. I'll add a ticket to do that

PR #151, opened 2017-03-14
add documentation for POST /caches/invalidate endpoint for apiV4

cpanato:
Implementation PR: https://github.com/mattermost/platform/pull/5756

### DIFF
--- a/v4/source/system.yaml
+++ b/v4/source/system.yaml
@@ -23,7 +23,7 @@
       description: |
         Recycle database connections by closing and reconnecting all connections to master and read replica databases.
         ##### Permissions
-        Must be authenticated and have the `PERMISSION_MANAGE_SYSTEM` permission.
+        Must be authenticated and have the `manage_system` permission.
       responses:
         '200':
           description: Database recycle successful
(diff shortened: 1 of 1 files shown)


v4/source/system.yaml
```diff
@@ -106,3 +106,20 @@
...
         '403':
           $ref: '#/responses/Forbidden'
+
+  '/caches/invalidate':
+    post:
+      tags:
+        - system
+      summary: Invalidate all the caches
+      description: |
+        Purge all the cache.
```
jwilander: Can we update this to 'Purge all the in-memory caches for the Mattermost server. This can have  a temporary negative effect on performance while the caches are re-populated.' ?

v4/source/system.yaml
```diff
@@ -106,3 +106,20 @@
...
+        ##### Permissions
+        Must be authenticated and have the `PERMISSION_MANAGE_SYSTEM` permission.
```
jwilander: Can you change the permission to be `manage_system`?

cpanato: I will update the others then because all are like this

cpanato:
@jwilander done

PR #158, opened 2017-03-15
Docs for APIv4 POST /channels/{channel_id}/members

saturninoabril:
This is document for endpoint APIv4 POST /channels/{channel_id}/members

### DIFF
--- a/v4/source/channels.yaml
+++ b/v4/source/channels.yaml
@@ -351,7 +351,41 @@
         '403':
           $ref: '#/responses/Forbidden'
 
-  '/channels/{channel_id}/members/ids':
+    post:
+      tags:
+        - channels
+      summary: Add a user
+      description: Add a user to the channel.
+      parameters:
+        - name: channel_id
+          in: path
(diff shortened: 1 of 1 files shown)


v4/source/channels.yaml
```diff
@@ -324,6 +324,40 @@
...
+      parameters:
+        - name: channel_id
+          in: path
+          description: The channel ID
+          required: true
+          type: string
+        - in: body
+          name: body
+          required: true
+          schema:
```
jwilander: Can we change this schema to just be a ref to ChannelMember?

saturninoabril: I'm not sure what you mean here. is it like below?
```
      parameters:
        - name: channel_id
          in: path
          description: The channel ID
          required: true
          type: string
        - in: body
          name: body
          required: true
          schema:
            $ref: '#/definitions/ChannelMember'
      responses:
        '200':
```

jwilander: Yes, just like that

PR #164, opened 2017-03-18
APIv4 POST /teams/{team_id}/members

saturninoabril:
This is document for endpoint APIv4 POST /teams/{team_id}/members.

### DIFF
--- a/v4/source/teams.yaml
+++ b/v4/source/teams.yaml
@@ -320,6 +320,54 @@
         '404':
           $ref: '#/responses/NotFound'
 
+    post:
+      tags:
+        - teams
+      summary: Add user to team
+      description: |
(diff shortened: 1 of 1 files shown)


v4/source/teams.yaml
```diff
@@ -276,6 +276,54 @@
...
+        - name: team_id
+          in: path
+          description: Team GUID
+          required: true
+          type: string
+        - name: user_id
+          in: query
+          description: The user_id of the user to add to the team.
+          required: false
+          type: string
```
jwilander: Can you change this to a ref to TeamMember and put it in the body?

saturninoabril: sure and sorry to miss this when I updated the endpoint.

PR #166, opened 2017-03-21
Document POST /users/search endpoint for APIv4

jwilander:


jwilander:
+1 on @hmhealey's changes

PR #167, opened 2017-03-21
add documentation for GET /users/{user_id}/status - user status 

cpanato:
implementation PR: https://github.com/mattermost/platform/pull/5824

### DIFF
--- a/v4/source/status.yaml
+++ b/v4/source/status.yaml
@@ -12,5 +12,7 @@
           description: User status retrieval successful
           schema:
             type: '#/definitions/Status'
+        '400':
+          $ref: '#/responses/BadRequest'
         '401':
           $ref: '#/responses/Unauthorized'
--- a/v4/source/status.yaml
+++ b/v4/source/status.yaml
@@ -1,0 +1,16 @@
+  /users/{user_id}/status:
+    get:
+      tags:
+        - status
+      summary: Get user status
+      description: |
+        Get user status by id from the server.
+        ##### Permissions
+        Must be authenticated.
+      responses:
+        '200':
+          description: User status retrieval successful
+          schema:
+            type: '#/definitions/Status'
+        '401':
+          $ref: '#/responses/Unauthorized'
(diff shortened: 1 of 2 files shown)


v4/source/status.yaml
```diff
@@ -0,0 +1,14 @@
...
+      summary: Get user status by id
+      description: |
+        Get user status by id from the server.
+        ##### Permissions
+        No permission required.
+      responses:
+        '200':
+          description: User status retrieval successful
+          schema:
+            type: string
```
jwilander: Can you update this to be a ref to the Status object? Might need to create it in definitions if it doesn't exist

v4/source/status.yaml
```diff
@@ -0,0 +1,14 @@
+  /users/{user_id}/status:
+    get:
+      tags:
+        - status
+      summary: Get user status by id
```
jwilander: Can we simplify this to just `Get user status`?

v4/source/status.yaml
```diff
@@ -0,0 +1,14 @@
...
+      description: |
+        Get user status by id from the server.
+        ##### Permissions
+        No permission required.
```
jwilander: For permission can we say `Must be authenticated.`?

v4/source/status.yaml
cpanato: done

v4/source/status.yaml
cpanato: done

v4/source/status.yaml
cpanato: done

v4/source/status.yaml
```diff
@@ -0,0 +1,16 @@
...
+      description: |
+        Get user status by id from the server.
+        ##### Permissions
+        Must be authenticated.
+      responses:
+        '200':
+          description: User status retrieval successful
+          schema:
+            type: '#/definitions/Status'
+        '401':
```
jwilander: Sorry, one last thing, can you add a 400 BadRequest response as well?

PR #172, opened 2017-03-22
PLT-2713 Added ability for admins to list users not in any team

hmhealey:

PR #182, opened 2017-03-24
PLT-6083: API to get users not in a team.

grundleborg:
API to get users not in a team.

https://mattermost.atlassian.net/browse/PLT-6083

### DIFF
--- a/v4/source/users.yaml
+++ b/v4/source/users.yaml
@@ -68,6 +68,10 @@
           in: query
           description: The ID of the team to get users for.
           type: string
+        - name: not_in_team
+          in: query
+          description: The ID of the team to exclude users for. Must not be used with "in_team" query parameter.
+          type: string
         - name: in_channel
           in: query
           description: The ID of the channel to get users for.


jwilander:
@hmhealey just added this ability for v4 in #172 and is going to hit that endpoint in the webapp using a custom driver function so he can use it in 3.8. Wondering if that can work for what you need as well?

hmhealey:
@jwilander Mine is different. It's an API for users not in any team. @grundleborg could do a similar method of just hooking into an APIv4 call though

jwilander:
Ah right, this is for not in a specific team not for users without any team.

Yes, preferably we just add this to the v4 endpoint instead of having to add a new v3 endpoint

hmhealey:
@grundleborg Also, if you end up just writing the API v4 version and calling it with the v3 client, make sure you remember that the v4 API returns a list of users instead of a map. It took me a fair bit of debugging to realize that I was merging an array and a map in the user store which caused all sorts of problems

grundleborg:
@hmhealey @jwilander updated. How's this look?

PR #189, opened 2017-03-28
Fix wrong return code

senk:
Wrong return code: 200 OK instead of 201 Created

TrivialFix


jwilander:
Actually this is a bug in the code, not the documentation. That should return a 201. @senk would you be interested in submitting a fix to the platform repo?

PR #190, opened 2017-03-28
APIv4 GET /teams/{team_id}/channels/deleted

senk:
Add Endpoint to fetch deleted channels

### DIFF
--- a/v4/source/channels.yaml
+++ b/v4/source/channels.yaml
@@ -166,7 +166,7 @@
         '200':
           description: Channel retrieval successful
           schema:
-            $ref: '#/definitions/Channel' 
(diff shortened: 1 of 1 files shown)


v4/source/channels.yaml
```diff
@@ -306,6 +306,37 @@
         '404':
           $ref: '#/responses/NotFound'
 
+  '/teams/{team_id}/channels/deleted':
+    get:
+      tags:
+        - teams
+      summary: Get deleted channels
+      description: |
+        Get a page of deleted channels on a team based on query string parameters - team_id.
```
jwilander: `team_id` is actually in the path, not the query string

senk:
@jwilander changed it

PR #192, opened 2017-03-28
Document PUT /channels/{channel_id}/members/{user_id}/notify_props for APIv4

jwilander:


jwilander:
Merged immediately to fix some errors.

PR #195, opened 2017-03-30
add documentation from POST /teams/{team_id}/import

cpanato:
implementation pr: https://github.com/mattermost/platform/pull/5920

### DIFF
--- a/v4/source/teams.yaml
+++ b/v4/source/teams.yaml
@@ -784,3 +784,45 @@
           $ref: '#/responses/Unauthorized'
         '403':
           $ref: '#/responses/Forbidden'
+
+  '/teams/{team_id}/import':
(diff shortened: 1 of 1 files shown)


v4/source/teams.yaml
```diff
@@ -644,3 +644,45 @@
...
+          type: string
+      responses:
+        '200':
+          description: File attached with the results of the import
+          schema:
+            type: file
+        '400':
+          $ref: '#/responses/BadRequest'
+        '401':
+          $ref: '#/responses/Unauthorized'
```
jwilander: I think we also need 403 forbidden here, right?

cpanato: done and rebased

PR #197, opened 2017-03-31
add documentation for PUT /users/{user_id}/status

cpanato:
Implementation pr: https://github.com/mattermost/platform/pull/5928

### DIFF
--- a/v4/source/status.yaml
+++ b/v4/source/status.yaml
@@ -1,4 +1,4 @@
-  /users/{user_id}/status:
+/users/{user_id}/status:
     get:
       tags:
         - status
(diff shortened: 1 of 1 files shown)


v4/source/status.yaml
```diff
@@ -16,3 +21,44 @@
...
+              user_id:
+                type: string
+              status:
+                type: string
+              manual:
+                type: boolean
+              last_activity_at:
+                type: integer
+              active_channel:
+                type: string
```
jwilander: For the properties can we only include `status`? The endpoint only uses that one so we should make sure that the documentation doesn't imply you can update the other fields

v4/source/status.yaml
```diff
@@ -16,3 +21,44 @@
           $ref: '#/responses/BadRequest'
         '401':
           $ref: '#/responses/Unauthorized'
+    put:
+      tags:
+        - status
+      summary: Update user status
+      description: |
+        Update user status.
```
jwilander: Can we change this to `Manually set a user's status. When setting a user's status, the status will remain that value until set "online" again, which will return the status to being automatically updated based on user activity.`

v4/source/status.yaml
cpanato: done

v4/source/status.yaml
cpanato: done

PR #218, opened 2017-04-19
add documentation for POST /logs

cpanato:
PR: https://github.com/mattermost/platform/pull/6143

### DIFF
--- a/v4/source/system.yaml
+++ b/v4/source/system.yaml
@@ -230,6 +230,41 @@
         '403':
           $ref: '#/responses/Forbidden'
 
+    post:
+      tags:
+        - system
+      summary: Add log message
+      description: |
+        Add log messages to the server logs.
+        ##### Permissions
+        Must have `manage_system` permission or if `ServiceSettings.EnableDeveloper` in the
+        config file is set to `true` then any user can use this endpoint.
+      parameters:
+        - in: body
+          name: body
+          required: true
+          schema:
+            type: object
+            required:
+              - level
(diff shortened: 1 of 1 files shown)


v4/source/system.yaml
```diff
@@ -230,6 +230,40 @@
         '403':
           $ref: '#/responses/Forbidden'
 
+    post:
+      tags:
+        - system
+      summary: Post logs
```
jwilander: Can we update this to "Add log message"?

v4/source/system.yaml
```diff
@@ -230,6 +230,40 @@
...
+      description: |
+        Post logs to the server logs.
```
jwilander: And this to "Add log messages to the server logs."?

v4/source/system.yaml
```diff
@@ -230,6 +230,40 @@
...
+        ##### Permissions
+        Must have `manage_system` permission.
```
jwilander: And here maybe mention that if `ServiceSettings.EnableDeveloper` is set to true in the config file, then any user can use this endpoint

v4/source/system.yaml
cpanato: done

v4/source/system.yaml
cpanato: done

v4/source/system.yaml
cpanato: done.

PR #219, opened 2017-04-19
APIv4 POST /channels/group

saturninoabril:
This is endpoint documentation for `APIv4 POST /channels/group`.

### DIFF
--- a/v4/source/channels.yaml
+++ b/v4/source/channels.yaml
@@ -80,6 +80,36 @@
         '403':
           $ref: '#/responses/Forbidden'
 
+  /channels/group:
+    post:
+      tags:
+        - channels
+      summary: Create a group message channel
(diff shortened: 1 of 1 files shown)


v4/source/channels.yaml
```diff
@@ -80,6 +80,36 @@
         '403':
           $ref: '#/responses/Forbidden'
 
+  /channels/group:
+    post:
+      tags:
+        - channels
+      summary: Create a group message channel
+      description: |
+        Create a new group message channel to group of users.
```
jwilander: Can we add a note along the lines of "If the logged in user's id is not included in the list, it will be appended to the end."?

saturninoabril:
ok, got it. see updated PR. thanks!

PR #222, opened 2017-04-23
APIv4 POST /commands/execute

saturninoabril:
This is endpoint documentation for `APIv4 POST /commands/execute`.

### DIFF
--- a/v4/source/commands.yaml
+++ b/v4/source/commands.yaml
@@ -282,3 +282,43 @@
 
             // RegenCommandToken
             newToken, resp := Client.RegenCommandToken(<COMMANDID>)
+
+  /commands/execute:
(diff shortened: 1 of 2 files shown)


v4/source/commands.yaml
```diff
@@ -204,3 +204,43 @@
...
+              - command
+            properties:
+              channel_id:
+                type: string
+                description: Channel Id where the command will execute
+              command:
+                type: string
+                description: The slash command to execute
+      responses:
+        '201':
```
jwilander: Can we switch this to `200`?

jwilander:
@saturninoabril I think we forgot about this one, would you be able to rebase and see my one comment?

PR #223, opened 2017-04-24
Docs changes for plt-6331

crspeller:


v4/source/users.yaml
```diff
@@ -941,14 +941,10 @@
...
             type: object
             required:
-              - userId
-              - hashId
+              - token
             properties:
-              user_id:
-                description: The user GUID
-                type: string
               hash_id:
```
jwilander: Whoops I think you need to change `hash_id` to `token` here

crspeller:
@jwilander Fixed!

PR #224, opened 2017-04-24
PLT-4457 Added API to get multiple users by their usernames

hmhealey:
This is needed for getting any users that have been at mentioned in a post that are not already loaded by the client

### DIFF
--- a/v4/source/users.yaml
+++ b/v4/source/users.yaml
@@ -128,6 +128,36 @@
         '401':
           $ref: '#/responses/Unauthorized'
 
+  /users/usernames:
+    post:
+      tags:
+        - users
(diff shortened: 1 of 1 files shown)


v4/source/users.yaml
```diff
@@ -128,6 +128,36 @@
         '401':
           $ref: '#/responses/Unauthorized'
 
+  /users/names:
```
jwilander: I know it's kind of redundant but I think we should use the actual field name so `users/usernames`. I think we also use `username` for the get by username endpoint

hmhealey: Oh, so we do. I specifically looked for that and I thought it was name. I'll fix that

PR #227, opened 2017-04-27
APIv4 POST /channels/{channel}/restore

senk:
Add Endpoint to restore deleted channels


jwilander:
Looks good @senk

jwilander:
Yep, thanks @senk

PR #230, opened 2017-04-29
CreateUser Code Sample

Zaicon:
CreateUser code sample:

http://i.imgur.com/UpZ1xgl.png


jwilander:
Sorry, I kind of let this one languish. We should probably get the other documentation for the C# driver setup before adding code samples.

Closing for now, @Zaicon we can reconnect on pre-release if you're interested in continuing work on the driver

PR #231, opened 2017-04-30
status.yaml: indent path

dagit:
It looks like there is a typo in `status.yaml`. At least, my yaml parser was failing and it looks like in other files the paths are always indented 2 spaces and this one wasn't. This fixes that.


jwilander:
Nice catch, thanks @dagit !

PR #233, opened 2017-05-03
Add missing WebSocket events

jasonblais:
Add missing WebSocket events from https://github.com/mattermost/platform/blob/master/model/websocket_message.go

### DIFF
--- a/v3/source/introduction.yaml
+++ b/v3/source/introduction.yaml
@@ -210,16 +210,25 @@ tags:
       - channel_created
       - channel_deleted
       - direct_added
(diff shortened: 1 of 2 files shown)


jasonblais:
Looks like this would cover https://github.com/mattermost/mattermost-api-reference/pull/232 as well

jasonblais:
I added in both v3/source../.. and v4/source../..

Or is there somewhere else I should add it to?

jwilander:
Whoops, guess I tunnel visioned a bit

PR #235, opened 2017-05-08
add documentation for POST /opengraph

cpanato:
code pr: https://github.com/mattermost/platform/pull/6343

### DIFF
--- a/v4/source/opengraph.yaml
+++ b/v4/source/opengraph.yaml
@@ -1,0 +1,17 @@
+  /opengraph:
+    post:
+      tags:
+        - opengraph
+      summary: Get open graph metadata for url
+      description: |
+        Get Open Graph Metadata for a specif URL. Use the Open Graph protocol to get some generic metadata about a URL. Used for creating link previews.
+      __Minimum server version__: 3.1
+        ##### Permissions
+        No permission required but must be logged in.
+      responses:
+        '200':
+          description: Open Graph retrieval successful
+          schema:
+            $ref: '#/definitions/OpenGraph'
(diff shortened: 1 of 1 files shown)


v4/source/opengraph.yaml
```diff
@@ -0,0 +1,16 @@
+  /opengraph:
+    post:
+      tags:
+        - opengraph
+      summary: Get Open Graph Metadata for a specif URL
```
jwilander: Can we change this to `Get open graph metadata for url`?

v4/source/opengraph.yaml
```diff
@@ -0,0 +1,16 @@
...
+      description: |
+        Get Open Graph Metadata for a specif URL.
```
jwilander: And this to `Use the Open Graph protocol to get some generic metadata about a URL. Used for creating link previews.`

v4/source/opengraph.yaml
```diff
@@ -0,0 +1,16 @@
...
+        ##### Permissions
+        No permission required.
```
jwilander: I think you need to be logged in right?

v4/source/opengraph.yaml
cpanato: done

v4/source/opengraph.yaml
cpanato: done, hope put the `Minimum`in the right place

v4/source/opengraph.yaml
cpanato: done

PR #238, opened 2017-05-10
PLT-4608 Added APIs for querying job status

hmhealey:
These are the first APIs needed for the job server. As far as we have designed right now, these are the only APIs needed specifically for the job server as configuration changes will be handled by modifying config.json through existing routes.

### Ticket Link
https://mattermost.atlassian.net/browse/PLT-6408

### DIFF
--- a/v4/source/jobs.yaml
+++ b/v4/source/jobs.yaml
@@ -4,9 +4,10 @@
         - jobs
       summary: Get the status of a given job.
       description: |
-          Gets the status of a single job,
-          ##### Permissions
-          Must have `manage_system` permission.
+        Gets the status of a single job.
+        __Minimum server version: 3.10__
+        ##### Permissions
+        Must have `manage_system` permission.
       parameters:
         - name: job_id
           in: path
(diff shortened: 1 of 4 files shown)


v4/source/jobs.yaml
```diff
@@ -0,0 +1,67 @@
...
+          $ref: '#/responses/NotFound'
+
+  '/jobs/type/{type}/statuses':
+    get:
+      tags:
+        - jobs
+      summary: Get the statuses of jobs of the given type.
+      description: |
+        Get a page of statuses of jobs of the given type. Use the query parameters to modify the behaviour of this endpoint.
+        ##### Permissions
```
jwilander: And need the minimum server version here as well

v4/source/jobs.yaml
```diff
@@ -0,0 +1,67 @@
+  '/jobs/{job_id}/status':
+    get:
+      tags:
+        - jobs
+      summary: Get the status of a given job.
+      description: |
+          Gets the status of a single job,
```
jwilander: Oh actually can you add `__Minimum server version: 3.10__` on a new line? We need to start doing that for everything after 3.9

Also you have typo `,` instead of `.`

PR #240, opened 2017-05-15
[PLT-6440] add code samples for Channel

cpanato:


jwilander:
Thanks for this @cpanato, I'll review soon just trying to finish some things :)

cpanato:
No rush @jwilander :)

v4/source/channels.yaml
```diff
@@ -146,6 +177,17 @@
...
         '404':
           $ref: '#/responses/NotFound'
+      x-code-samples:
+        - lang: 'Go'
+          source: |
+            import "github.com/mattermost/platform/model"
+            Client := model.NewAPIv4Client("https://your-mattermost-url.com")
+            Client.Login("email@domain.com", "Password1")
+
+            input := []string{<ID OF CHANNEL1>, <ID OF CHANNEL2>, ...}
```
jwilander: Maybe change `input` to be `channelIds`? Just makes it a bit clearer I think

v4/source/channels.yaml
```diff
@@ -304,6 +375,26 @@
...
+            patch := &model.ChannelPatch{
+              Name:        new(string),
+              DisplayName: new(string),
+              Header:      new(string),
+              Purpose:     new(string),
+            }
+            *patch.Name = model.NewId()
+            *patch.DisplayName = model.NewId()
+            *patch.Header = model.NewId()
+            *patch.Purpose = model.NewId()
```
jwilander: Maybe instead of `model.NewId()` we could do `"somename"`, `"somedisplayname"`, etc. I think if someone doesn't know what `model.NewId()` is that could cause confusion

v4/source/channels.yaml
cpanato: done

v4/source/channels.yaml
cpanato: done

PR #246, opened 2017-06-02
Remove creator_id from required fields for command

94117nl:
Remove creator_id from required fields for create command, add more fields to request sample


jwilander:
Thanks @94117nl !

PR #247, opened 2017-06-04
Update ping health and add code examples

cpanato:


v4/source/system.yaml
```diff
@@ -2,18 +2,31 @@
     get:
       tags:
         - system
-      summary: Ping the Server
+      summary: Check system health
       description: |
-        Receive a simple ping response.
+        Check if the system is health or unhealthy based on the configuration setting `GoRoutineHealthThreshold`.
```
jwilander: Can we update this to "Check if the server is up and healthy based on the configuration setting \`GoRoutineHealthThreshold\`. If \`GoRoutineHealthThreshold\` and the number of goroutines on the server exceeds that threshold the server is considered unhealthy. If \`GoRoutineHealthThreshold\` is not set or the number of goroutines is below the threshold the server is considered healthy."

v4/source/system.yaml
```diff
@@ -2,18 +2,31 @@
...
+        __Minimum server version__: 3.10
         ##### Permissions
-        No permissions required.
+        Must have `manage_system` permission
```
jwilander: There is no permission required

v4/source/system.yaml
```diff
@@ -2,18 +2,31 @@
...
       responses:
         '200':
-          description: Ping successful
+          description: Status of the system
           schema:
```
jwilander: The schema for 200 didn't change, I think this needs to be under '500'

jwilander:
Can you also submit the code samples as a separate PR?

v4/source/system.yaml
cpanato: done

v4/source/system.yaml
cpanato: done

v4/source/system.yaml
cpanato: done

PR #250, opened 2017-06-07
[PLT-6440] add code samples for sytem api

cpanato:
the Ping endpoint sample is in another PR.

### DIFF
--- a/v4/source/system.yaml
+++ b/v4/source/system.yaml
@@ -30,7 +30,7 @@
             Client.Login("email@domain.com", "Password1")
 
             // GetPing
-            status, err := Client.GetPing()
+            status, resp := Client.GetPing()
 
   '/database/recycle':
     post:
@@ -56,8 +56,7 @@
             Client := model.NewAPIv4Client("https://your-mattermost-url.com")
             Client.Login("email@domain.com", "Password1")
 
-            // DatabaseRecycle
-            flag, err := Client.DatabaseRecycle()
+            ok, resp := Client.DatabaseRecycle()
 
   '/email/test':
     post:
@@ -86,7 +85,7 @@
             Client.Login("email@domain.com", "Password1")
 
             // TestEmail
-            flag, err := Client.TestEmail()
+            ok, resp := Client.TestEmail()
 
   '/config':
     get:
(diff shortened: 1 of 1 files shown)


v4/source/system.yaml
```diff
@@ -87,6 +117,21 @@
...
+            Client := model.NewAPIv4Client("https://your-mattermost-url.com")
+            Client.Login("email@domain.com", "Password1")
+
+            // GetConfig
+            config, err := Client.GetConfig()
+
+            config.TeamSettings.SiteName = "MyFancyName"
+
+            // UpdateConfig
+            updatedConfig, err := Client.UpdateConfig()
```
jwilander: This should be `Client.UpdateConfig(config)`, right?

v4/source/system.yaml
```diff
@@ -31,6 +31,16 @@
...
+      x-code-samples:
+        - lang: 'Go'
+          source: |
+            import "github.com/mattermost/platform/model"
+
+            Client := model.NewAPIv4Client("https://your-mattermost-url.com")
+            Client.Login("email@domain.com", "Password1")
+
+            // DatabaseRecycle
+            flag, err := Client.DatabaseRecycle()
```
jwilander: Can we use `ok` instead of `flag`? I think that's more standard for Go

In all the other samples as well

jwilander: Also maybe remove the `// DatabaseRecycle` comments, not sure if they're much help since it's just the name of the function

cpanato: done, sorry for the long delay

v4/source/system.yaml
cpanato: yes true! done.

v4/source/system.yaml
```diff
@@ -48,6 +48,15 @@
...
           $ref: '#/responses/Forbidden'
+      x-code-samples:
+        - lang: 'Go'
+          source: |
+            import "github.com/mattermost/platform/model"
+
+            Client := model.NewAPIv4Client("https://your-mattermost-url.com")
+            Client.Login("email@domain.com", "Password1")
+
+            ok, err := Client.DatabaseRecycle()
```
jwilander: The `err` should be `resp` here and in the other examples too

cpanato: done

PR #251, opened 2017-06-07
Add docs for APIv4 permanent delete

ZJvandeWeg:
Needed for https://github.com/mattermost/platform/pull/6604

### DIFF
--- a/v4/source/teams.yaml
+++ b/v4/source/teams.yaml
@@ -168,6 +168,12 @@
           description: Team GUID
           required: true
           type: string
+        - name: permanent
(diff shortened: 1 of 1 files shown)


v4/source/teams.yaml
```diff
@@ -168,6 +168,12 @@
           description: Team GUID
           required: true
           type: string
+        - name: permanent
+          in: body
```
jwilander: I think this needs to be `in: query`

ZJvandeWeg:
@jwilander Fixed 😄

PR #262, opened 2017-06-29
PLT-6474: Elasticsearch test config endpoint.

grundleborg:
Admin endpoint for testing elasticsearch config.

https://mattermost.atlassian.net/browse/PLT-6474

### DIFF
--- a/v4/source/elasticsearch.yaml
+++ b/v4/source/elasticsearch.yaml
@@ -1,0 +1,20 @@
+  /elasticsearch/test:
+    post:
+      tags:
+        - elasticsearch
+      summary: Test Elasticsearch configuration
+      description: |
+        Test the current Elasticsearch configuration to see if the Elasticsearch server can be contacted successfully.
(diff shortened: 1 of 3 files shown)


v4/source/elasticsearch.yaml
```diff
@@ -0,0 +1,19 @@
+  /elasticsearch/test:
+    post:
+      tags:
+        - elasticsearch
+      summary: Test Elasticsearch configuration
+      description: |
+        Test the current Elasticsearch configuration to see if the Elasticsearch server can be contacted successfully.
```
jwilander: Can you add `**Minimum server version: 4.0**` to the description on a new line?

jwilander: Actually maybe `**Minimum server version: unreleased**` or something would be better to start with

grundleborg: @jwilander done.

PR #267, opened 2017-07-06
PLT-5100 add documentation for getUsersStats

jsfeb26:


v4/source/users.yaml
```diff
@@ -1118,3 +1118,25 @@
         '501':
           $ref: '#/responses/NotImplemented'
 
+  /users/stats:
+    get:
+      tags:
+        - users
+      summary: Get users stats
+      description: |
+        Get users stats on the system.
```
jwilander: Could you update this to say:
```
Get statistics for users on the system.

**Minimum server version:** unreleased
```

Once the code gets merged in we can update the minimum server version to be whatever the next one is

v4/source/users.yaml
```diff
@@ -1118,3 +1118,25 @@
...
+          description: Users stats retrieval successful
+          schema:
+            $ref: '#/definitions/UsersStats'
+        '400':
+          $ref: '#/responses/BadRequest'
+        '401':
+          $ref: '#/responses/Unauthorized'
+        '403':
+          $ref: '#/responses/Forbidden'
+        '404':
```
jwilander: I don't think this endpoint will ever return 404, mind removing this error response?

jwilander:
Closing due to inactivity, please re-open if you'd like to continue work on this

PR #269, opened 2017-07-10
PLT-7025: Team Import API should return JSON, not attachment.

grundleborg:
This helps fix PLT-7025 while also leaving the opportunity open to send back more structured data in future without breaking backwards compatibility.

@crspeller @jwilander what do you think of a) this returning JSON and b) the property name `results`?

### DIFF
--- a/v4/source/teams.yaml
+++ b/v4/source/teams.yaml
@@ -849,9 +849,12 @@
           type: string
       responses:
         '200':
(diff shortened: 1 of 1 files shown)


jwilander:
I think it's fine, it also leaves room to add other fields to the JSON response in the future without breaking changes

PR #270, opened 2017-07-11
PLT-6595: Jobs API.

grundleborg:
Proposed API for Jobs.

Needed to start on the system console UI for jobs in https://mattermost.atlassian.net/browse/PLT-6595

### DIFF
--- a/v4/source/definitions.yaml
+++ b/v4/source/definitions.yaml
@@ -1373,6 +1373,34 @@ definitions:
         type: integer
         description: The last time of update for the application
 
+  Job:
+    type: object
+    properties:
+      id:
+        type: string
+        description: The unique id of the job
+      type:
+        type: string
+        description: The type of job
+      create_at:
+        type: integer
+        description: The time at which the job was created
+      start_at:
+        type: integer
+        description: THe time at which the job was started
+      last_activity_at:
+        type: integer
+        description: The last time at which the job had activity
+      status:
+        type: string
+        description: The status of the job
+      progress:
+        type: integer
+        description: The progress (as a percentage) of the job
+      data:
+        type: object
+        description: A freeform data field containing additional information about the job
+
 externalDocs:
   description: Find out more about Mattermost
   url: 'https://about.mattermost.com'
--- a/v4/source/definitions.yaml
+++ b/v4/source/definitions.yaml
@@ -1387,7 +1387,7 @@ definitions:
         description: The time at which the job was created
       start_at:
         type: integer
-        description: THe time at which the job was started
+        description: The time at which the job was started
       last_activity_at:
         type: integer
         description: The last time at which the job had activity
--- a/v4/source/introduction.yaml
+++ b/v4/source/introduction.yaml
@@ -303,6 +303,8 @@ tags:
     description: Endpoints for configuring and interacting with LDAP.
   - name: system
     description: General endpoints for interating with the server, such as configuration and logging.
+  - name: Jobs
+    description: Endpoints related to various background jobs that can be run by the server or separately by jobservers.
   - name: OAuth
     description: Endpoints for configuring and interacting with Mattermost as an OAuth 2.0 service provider.
   - name: elasticsearch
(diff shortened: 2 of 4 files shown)


v4/source/definitions.yaml
```diff
@@ -1373,6 +1373,34 @@ definitions:
...
+        description: The unique id of the job
+      type:
+        type: string
+        description: The type of job
+      create_at:
+        type: integer
+        description: The time at which the job was created
+      start_at:
+        type: integer
+        description: THe time at which the job was started
```
jwilander: Typoe, `THe` -> `The`

v4/source/introduction.yaml
```diff
@@ -303,6 +303,8 @@ tags:
     description: Endpoints for configuring and interacting with LDAP.
   - name: system
     description: General endpoints for interating with the server, such as configuration and logging.
+  - name: Jobs
+    description: Endpoints related to various background jobs that can be run by the server or separately by jobservers.
```
jwilander: Should jobservers be two words?

v4/source/jobs.yaml
```diff
@@ -0,0 +1,171 @@
+  '/jobs':
+    get:
+      tags:
+        - jobs
+      summary: Get the jobs.
+      description: |
+        Get a page of jobs. Use the query parameters to modify the behaviour of this endpoint.
+        __Minimum server version: 4.1__
+        ##### Permissions
+        Must have `manage_system` permission.
```
jwilander: Maybe we should add new permissions for jobs since we'll be breaking it all out for advanced permissions anyways

v4/source/jobs.yaml
```diff
@@ -0,0 +1,171 @@
...
+          $ref: '#/responses/Unauthorized'
+        '403':
+          $ref: '#/responses/Forbidden'
+
+    post:
+      tags:
+      - jobs
+      summary: Create a new job.
+      description: |
+        Create a new job.
```
jwilander: I realize it probably doesn't exist yet but later we should link to documentation about how jobs work

v4/source/jobs.yaml
```diff
@@ -0,0 +1,171 @@
...
+            type: object
+            required:
+              - type
+            properties:
+              type:
+                type: string
+                description: The type of job to create
+              data:
+                type: object
+                description: An object containing any additional data required for this job type
```
jwilander: Same as above, should link documentation for the different job types and what data to include for them when we have it

v4/source/jobs.yaml
```diff
@@ -0,0 +1,171 @@
...
+          $ref: '#/responses/BadRequest'
+        '401':
+          $ref: '#/responses/Unauthorized'
+        '403':
+          $ref: '#/responses/Forbidden'
+        '404':
+          $ref: '#/responses/NotFound'
+
+  '/jobs/{job_id}/cancel':
+    get:
```
jwilander: Does this delete the job? Or is cancel a separate action? If it's performing a delete we should use DELETE, otherwise we should use POST (we use that for actions not covered by the HTTP verbs we're using)

v4/source/jobs.yaml
grundleborg: 0/5 on this, but wondering if there's any point adding new ones at this stage, given what's wanted might change when speccing advanced permissions.

v4/source/jobs.yaml
grundleborg: OK. Will do this once such documentation exists.

v4/source/jobs.yaml
grundleborg: :+1:

v4/source/jobs.yaml
grundleborg: It doesn't delete the job (jobs can't be deleted through the API). It just sets it to have a status of "request_canceled". It's the only change that anyone other than the worker is allowed to make to a Job once it's been created, so having a whole `update` endpoint seemed inappropriate. I'll change it to a POST though.

v4/source/jobs.yaml
jwilander: Yeah that's fine, we'll already to have to rework a bunch of permissions as is

PR #272, opened 2017-07-14
Fixed user/{user_id}/teams return schema

thePanz:
See implementation: https://github.com/mattermost/platform/pull/5296


jwilander:
Thanks @thePanz, nice catch

PR #277, opened 2017-08-03
PLT-6597: Add search matches to posts search results.

grundleborg:
This adds the search matches to highlight to the data returned by the server.

For Elasicsearch, we need to return the "highlights" from the server like this, as it's impractical to reconstruct them client side when the system is quite complicated, and ES provides them anyway. For ordinary database search, I'm not sure whether it would make sense to also do this server side for a consistent API (I don't think it would be much work to move what we currently do to the server side in this case), or whether to leave it client side.

### DIFF
--- a/v4/source/definitions.yaml
+++ b/v4/source/definitions.yaml
@@ -207,6 +207,26 @@ definitions:
         additionalProperties:
           $ref: '#/definitions/Post'
 
+  PostListWithSearchMatches:
+    type: object
+    properties:
+      order:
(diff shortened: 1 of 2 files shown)


jwilander:
I think it would be nicer to move it all server-side just to be consistent, so instead of just a new API feature for Elasticsearch it's a new feature for search across both

grundleborg:
Closing this as HH is going to take over implementing this functionality.

PR #290, opened 2017-09-22
PLT-7705: Add a data retention policy endpoint.

grundleborg:
This adds an API endpoint to get the data retention policy. The purpose is to enable the mobile apps to ask the API "What are the cutoff timestamps before which we should remove all messages/files we have cached?".

https://mattermost.atlassian.net/browse/PLT-7705

### DIFF
--- a/v4/source/dataretention.yaml
+++ b/v4/source/dataretention.yaml
@@ -6,6 +6,8 @@
       description: |
         Gets the current data retention policy details from the server, including what data should be purged and the cutoff times for each data type that should be purged.
         __Minimum server version__: 4.3
+        ##### Permissions
+        Requires an active session but no other permissions.
       responses:
         '200':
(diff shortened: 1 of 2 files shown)


v4/source/dataretention.yaml
```diff
@@ -0,0 +1,17 @@
+  /data_retention/policy:
+    get:
+      tags:
+        - dataretention
+      summary: Get the data retention policy details.
+      description: |
+        Gets the current data retention policy details from the server, including what data should be purged and the cutoff times for each data type that should be purged.
+        __Minimum server version__: 4.3
```
jwilander: Can you add a permissions section:
```
##### Permissions
This is the required permission.
```

grundleborg: @jwilander done.

PR #295, opened 2017-10-16
[PLT-7794] Added user access token enable/disable endpoints

fraziern:
Adds documentation for the forthcoming server PR as part of https://github.com/mattermost/mattermost-server/issues/7583 and PLT-7794

A few points to note:
 * I wasn't sure what to put for the minimum server version, so feel free to provide input there
 * These endpoints currently rely on existing permissions rather than new ones
 * I created 2 endpoints rather than 1 unified "update token active status" endpoint because they do different things. Disabling a token also wipes out its related sessions.

### DIFF
--- a/v4/source/users.yaml
+++ b/v4/source/users.yaml
@@ -1272,9 +1272,9 @@
     post:
       tags:
         - users
-      summary: Temporarily disable a user access token
+      summary: Disable personal access token
       description: |
-        Disable a user access token and delete any sessions using the token. The token can be re-enabled using `/users/tokens/enable`.
+        Disable a personal access token and delete any sessions using the token. The token can be re-enabled using `/users/tokens/enable`.
 
         __Minimum server version__: 4.4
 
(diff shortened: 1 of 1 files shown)


jwilander:
* 4.4 is good 👍 
* Permissions are ok
* Two endpoints makes sense

v4/source/users.yaml
```diff
@@ -1268,3 +1268,74 @@
         '404':
           $ref: '#/responses/NotFound'
 
+  /users/tokens/disable:
+    post:
+      tags:
+        - users
+      summary: Temporarily disable a user access token
```
jwilander: Maybe shorten the summaries to `Disable personal access token` and `Enable personal access token`, they go into a sidebar so shorter the better

Also the documentation should refer to them as "personal access tokens", while in code they are "user access tokens"

fraziern:
Changes made.

jwilander:
Thanks @fraziern !

PR #297, opened 2017-10-24
Current password in /users/{user_id}/password only required if updating your own password

jasonblais:


jasonblais:
@jwilander please help with review

PR #298, opened 2017-10-25
Move some channel endpoints to be under channel section

jwilander:
Copy/paste error had them under teams accidentally.


Vaelor:
@jwilander Did this change not make it into the 4.4.0 release? The api documentation changed but nothing is in the release notes for 4.4.0 and it does not seem to work
https://docs.mattermost.com/administration/changelog.html#release-v4-4-0

PR #301, opened 2017-11-01
Add minimum server version for /users/{user_id}/sessions/revoke

jasonblais:


jasonblais:
Nevermind - it was `/users/{user_id}/sessions/revoke/all` that got added in v4.4 and that already mentions the minimum server version.

PR #315, opened 2017-12-23
[PLT-8173] updated documentation for creating incoming webhooks

lieut-data:
This documentation complements the changes to support https://github.com/mattermost/mattermost-server/pull/8002, exposing a Username and Profile Picture configuration for incoming webhooks to override the defaults.

See https://github.com/mattermost/mattermost-server/issues/7880 for the original ticket.

### DIFF
--- a/v4/source/webhooks.yaml
+++ b/v4/source/webhooks.yaml
@@ -26,10 +26,10 @@
               description:
                 type: string
                 description: The description for this incoming webhook
-              post_username:
+              username:
                 type: string
                 description: The username this incoming webhook will post as.
-              post_icon_url:
+              icon_url:
                 type: string
                 description: The profile picture this incoming webhook will use when posting.
       responses:
@@ -148,10 +148,10 @@
               description:
                 type: string
                 description: The description for this incoming webhook
-              post_username:
+              username:
                 type: string
                 description: The username this incoming webhook will post as.
-              post_icon_url:
+              icon_url:
                 type: string
                 description: The profile picture this incoming webhook will use when posting.
       responses:
--- a/v4/source/webhooks.yaml
+++ b/v4/source/webhooks.yaml
@@ -26,6 +26,12 @@
               description:
                 type: string
                 description: The description for this incoming webhook
+              post_username:
+                type: string
+                description: The username this incoming webhook will post as.
+              post_icon_url:
+                type: string
+                description: The profile picture this incoming webhook will use when posting.
       responses:
         '201':
           description: Incoming webhook creation successful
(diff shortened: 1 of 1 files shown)


v4/source/webhooks.yaml
```diff
@@ -26,6 +26,12 @@
               description:
                 type: string
                 description: The description for this incoming webhook
+              post_username:
```
lieut-data: See the complementing PRs, but I debated whether or not to just call this parameter `username` and `icon_url` vs. the `post_` prefixed versions here. I ended up picking the `Post` prefix to complement the `Post` in `EnablePostUsernameOverride`. Let me know if you agree!

enahum: what about override_username and override_icon_url?

lieut-data: Yes, I suppose that's another option as well. The configuration being called, "PostUsernameOverride", I had originally assumed that the entity being overridden was the "post username", until I found other code just using "override username", hence the uncertainty.

Your suggestion has the nice property of aligning with the parameter name used when posting to the webhook, making things at least consistent from the API perspective.

Let me know how you'd like to proceed and I'll make any necessary changes.

enahum: Yeah lets make the change ;)

lieut-data: Will do. Note that this will of course impact https://github.com/mattermost/mattermost-webapp/pull/514 and https://github.com/mattermost/mattermost-server/pull/8002, requiring a re-review. I'll ping all of these PRs once updated.

lieut-data: @enahum, looking more closely at the documentation again, I see that the parameter names used when posting a message via the incoming webhook are actually just `username` and `icon_url`. I think I had originally prefixed with `post_` when creating the integration so as to distinguish from some future model attribute, but I suspect that's not necessary.

Would you feel comfortable with converging on just `username` and `icon_url`, dispensing with both the `post_` and `override_` prefixes?

enahum: @jwilander thoughts? I’m thinking leaving it as override_ as we have that everywhere with the webhooks

lieut-data: Note that @crspeller kindly merged my approved reviews related to this change, so if we decide to change the name of this parameter, I'll open up new PRs as such.

jwilander: Yeah I'd prefer to converge on just `username` and `icon_url`. If you don't mind opening a new server PR to match, that would be great

lieut-data:
Looks like this build is failing on some newly expanded style checks (namely checking `*.js` instead of just `*.jsx`). I'm addressing some naming feedback from https://github.com/mattermost/mattermost-api-reference/pull/315 and will resolve as part of those changes.

lieut-data:
I've made the requested changes above, including updating the outstanding webapp PR and filing a new server PR, referenced above.

jwilander:
@enahum if this looks good to you, please approve

PR #316, opened 2018-01-02
PLT-8179: Add role get / update endpoints to API.

grundleborg:
First draft of API of phase 1 permissions APIs:
- Fetch roles individually.
- Fetch roles in bulk by name.
- Update roles to alter the permissions they grant.


v4/source/roles.yaml
```diff
@@ -0,0 +1,150 @@
+  '/roles/name/{role_id}':
+    get:
+      tags:
+        - roles
+      summary: Get a role
+      description: |
+        Get a role from the provided role id.
+        ##### Permissions
+        Requires an active session but no other permissions..
```
jwilander: Double period typo

v4/source/roles.yaml
```diff
@@ -0,0 +1,150 @@
...
+        '404':
+          $ref: '#/responses/NotFound'
+      x-code-samples:
+        - lang: 'Go'
+          source: |
+            import "github.com/mattermost/mattermost-server/model"
+            Client := model.NewAPIv4Client("https://your-mattermost-url.com")
+            Client.Login("email@domain.com", "Password1")
+
+            // GetChannel
```
jwilander: Comment is wrong here (and for the other examples), can probably just be removed

PR #318, opened 2018-01-04
[PLT-7793] Added /users/tokens endpoint

watadarkstar:
#### Summary

Adds a new endpoint `/users/tokens` to the documentation.

#### PR Links

- [Documentation PR](https://github.com/mattermost/mattermost-api-reference/pull/318)
- [Webapp PR](https://github.com/mattermost/mattermost-webapp/pull/496)
- [Server PR](https://github.com/mattermost/mattermost-server/pull/8038)
- [Redux PR](https://github.com/mattermost/mattermost-redux/pull/356)

#### Ticket Link

- [Jira](https://mattermost.atlassian.net/browse/PLT-7793)
- [Github](https://github.com/mattermost/mattermost-server/issues/7584#issuecomment-352808489)

#### Notes

`is_active` was missing from the definitions so I added it too.

#### Checklist
- [x] Copy `.yaml` from existing route
- [x] Add `/users/tokens/all`
- [x] Update tags, summary, description, permissions, params, responses
- [x] Test with `make build-v4` and copy into Swagger

### DIFF
--- a/v4/source/users.yaml
+++ b/v4/source/users.yaml
@@ -1197,7 +1197,7 @@
       description: |
         Get a list of user access tokens for a user. Does not include the actual authentication tokens. Use query paremeters for paging.
 
-        __Minimum server version__: 4.7
+        __Minimum server version__: 4.1
 
         ##### Permissions
         Must have `read_user_access_token` permission. For non-self requests, must also have the `edit_other_users` permission.
@@ -1231,7 +1231,7 @@
         '403':
           $ref: '#/responses/Forbidden'
   
-  /users/tokens/all:
+  /users/tokens:
     get:
       tags:
         - users
@@ -1239,7 +1239,7 @@
       description: |
         Get a list of all the user access tokens for all users. Does not include the actual authentication tokens. Use query paremeters for paging.
 
-        __Minimum server version__: 4.1
+        __Minimum server version__: 4.7
 
         ##### Permissions
         Must have `manage_system` permission.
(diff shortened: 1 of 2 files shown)


v4/source/users.yaml
```diff
@@ -1230,7 +1230,44 @@
           $ref: '#/responses/Unauthorized'
         '403':
           $ref: '#/responses/Forbidden'
+  
+  /users/tokens:
+    get:
+      tags:
+        - users
+      summary: Get all user access tokens for all users
```
jwilander: How about just 'Get user access tokens'?

v4/source/users.yaml
```diff
@@ -1230,7 +1230,44 @@
...
+      description: |
+        Get a list of all the user access tokens for all users. Does not include the actual authentication tokens. Use query paremeters for paging.
```
jwilander: And this to 'Get a page of user access tokens for users on the system. Does not include the actual authentication tokens. Use query parameters for paging.'

v4/source/users.yaml
watadarkstar: On phone but I think that's not correct. It should be: 'Get users' access tokens' if anything. One is getting all user access tokens for all users.

jwilander:
I think it make sense grammatically if 'user access token' is the object

watadarkstar:
@jwilander Fixed all your requests! 😄

PR #326, opened 2018-01-24
ICU-573 added encoded filename to api docs

stephenkiers:
Related to changes here : https://github.com/mattermost/mattermost-server/pull/8150

### DIFF
--- a/v4/source/files.yaml
+++ b/v4/source/files.yaml
@@ -30,7 +30,7 @@
           type: string
         - name: encoded_filename
           in: formData
-          description: An optional bool string, **true**, that tells the server if the filename was encoded before upload so that it can be decoded before saving **Minimum server version: 4.7**
+          description: An optional bool string, "true", that tells the server if the filename was encoded before upload so that it can be decoded before saving Minimum server version: 4.7
           required: false
           allowEmptyValue: true
           type: string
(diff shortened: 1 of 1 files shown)


hmhealey:
Adding @jwilander since he's master of the API. Any thoughts on whether or not we should use a query parameter for this instead of a field in the form data?

jwilander:
0/5 on form data vs query parameter in this case. If we already have form data, it's better to include this there as well IMO. The goal is to be easier on the API user and if they're already creating form data then they don't need to worry about adding query parameters as well but it's not a huge deal either way

v4/source/files.yaml
```diff
@@ -28,6 +28,12 @@
           required: false
           allowEmptyValue: true
           type: string
+        - name: encoded_filename
+          in: formData
+          description: An optional bool string that tells the server if the filename was encoded before upload so that it can be decoded before saving
```
jwilander: In the description can we add a note for the minimum server version this will be included in? Since we don't version the reference with each server release

Just a note like this `**Minimum server version: 4.7**` (assuming it'll be in for 4.7)

stephenkiers: Markdown doesn't seem to be supported. I simply added the message

PR #327, opened 2018-01-30
ABC-79: Add channel autocomplete endpoint

ccbrown:
See https://github.com/mattermost/mattermost-server/pull/8163

### DIFF
--- a/v4/source/channels.yaml
+++ b/v4/source/channels.yaml
@@ -588,6 +588,42 @@
         '404':
           $ref: '#/responses/NotFound'
 
+  '/teams/{team_id}/channels/autocomplete':
+    post:
+      tags:
+        - channels
+      summary: Autocomplete channels
+      description: |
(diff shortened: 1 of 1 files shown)


v4/source/channels.yaml
```diff
@@ -588,6 +588,42 @@
...
+          description: Team GUID
+          required: true
+          type: string
+        - name: name
+          in: query
+          description: Name or display name
+          required: true
+          type: string
+      responses:
+        '201':
```
jwilander: Should be a 200 I think

v4/source/channels.yaml
```diff
@@ -588,6 +588,42 @@
         '404':
           $ref: '#/responses/NotFound'
 
+  '/teams/{team_id}/channels/autocomplete':
+    post:
```
jwilander: Should be a GET, right?

ccbrown:
@jwilander Derp. Fixed.

PR #329, opened 2018-02-03
MM-9274- Added sort=status querystring option on Users

stephenkiers:


v4/source/users.yaml
```diff
@@ -105,7 +105,16 @@
...
-            Can be "", "last_activity_at" or "create_at". When left blank sorting is done by username. Must be used in conjunction with `in_team` and no other query parameters (exlcuding the paging parameters). **Minimum server version: 4.0**
+            Sort is only available in conjunction with certain options, exlcuding the paging parameters
+
+            ##### `in_team`
+            Can be "", "last_activity_at" or "create_at".
+            When left blank sorting is done by username.
+            __Minimum server version__: 4.0
+            ##### `in_channel`
+            Can be "", "status".
+            When left blank sorting is done by username. `status` will sort by User's current status (Online, Away, Offline, DND), the by Username.
```
jwilander: Typo here `the by Username` -> `then by Username`

v4/source/users.yaml
```diff
@@ -105,7 +105,16 @@
         - name: sort
           in: query
           description: |
-            Can be "", "last_activity_at" or "create_at". When left blank sorting is done by username. Must be used in conjunction with `in_team` and no other query parameters (exlcuding the paging parameters). **Minimum server version: 4.0**
+            Sort is only available in conjunction with certain options, exlcuding the paging parameters
```
hmhealey: I'm not sure I understand what this means. Can it or can't it be used with paging?

stephenkiers: plus the spelling (which was inherited). I will attempt to reword to be clearer

Issue #330, opened 2018-02-06
Update configuration misses parameters from documentation


Vaelor:
https://api.mattermost.com/#tag/system%2Fpaths%2F~1config%2Fget has no documented parameters

According to:
https://github.com/mattermost/mattermost-server/blob/master/api4/system.go#L113
it seems you need to drop a config object (not sure if a whole config?) into the body when doing this action, this is missing from the documentation.

PR #331, opened 2018-02-08
Add reference to config definition (#330)

Vaelor:


jwilander:
Nice catch, thanks @Vaelor !

PR #332, opened 2018-02-12
ABC-89 Add note about searching posts from users and in channels

jwilander:


jasonblais:
Minor tweak for consistency, otherwise +1

PR #335, opened 2018-02-15
MM-9556 Added documentation for new file upload route

hmhealey:
This is for the new file upload API required for the iOS share extension. Since it's documentation, it's pretty self-explanatory

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-9556

### DIFF
--- a/v4/source/files.yaml
+++ b/v4/source/files.yaml
@@ -10,6 +10,9 @@
         client_ids defined in the FormData, or it can be a request with the channel_id and filename
         defined as query parameters with the contents of a single file in the body of the request.
 
+        Only multipart/form-data requests are supported by server versions up to and including 4.7.
+        Server versions 4.8 and higher support both types of requests.
+
         ##### Permissions
         Must have `upload_file` permission.
       consumes:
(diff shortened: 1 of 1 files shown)


v4/source/files.yaml
```diff
@@ -5,29 +5,45 @@
       summary: Upload a file
       description: |
         Uploads a file that can later be attached to a post.
+
+        This request can either be a multipart/form-data request with a channel_id, files and optional
+        client_ids defined in the FormData, or it can be a request with the channel_id and filename
+        defined as query parameters with the contents of a single file in the body of the request.
```
enahum: isn't this alternative request type suppose to be of content type `application/octet-stream` ?

hmhealey: It can be any type. The server doesn't really care what it is as long as it's not multipart

jwilander: I'm 0/5 here, it's pretty common not to send a content-type when you're just sending bytes

v4/source/files.yaml
```diff
@@ -5,29 +5,45 @@
       summary: Upload a file
       description: |
         Uploads a file that can later be attached to a post.
+
+        This request can either be a multipart/form-data request with a channel_id, files and optional
```
jwilander: Actually, can you add a note here about the new functionality only being available in 4.8+ servers?

PR #338, opened 2018-03-01
Fix "users/status/ids" API documentation

thePanz:


v4/source/status.yaml
```diff
@@ -60,14 +60,22 @@
...
-    get:
+    post:
       tags:
         - status
       summary: Get user statuses by id
       description: |
         Get a list of user statuses by id from the server.
         ##### Permissions
         Must be authenticated.
+      parameters:
```
cpanato: Hello @thePanz thanks for this update
just a small nit

can you change the following parameters to fix the build issue and also the fix the definition

 - `name` should be `post`
 - in the `schema` the items should be:
  ```
  items:
    type: string
  ```

the whole `parameters` section:
```
parameters:
    - name: post
       in: body
       description: List of user ids to fetch
       required: true
       schema:
           type: array
           items:
             type: string
```

thePanz: Updated the MR with your comments :)

cpanato:
cool! many thanks, just a very small request, can you please squash the commits into one? thanks so much again.

thePanz:
@cpanato I'll work on that during the next days. 
btw: can you check if github has a "merge and squash" option for PRs? it would do it too :)

cpanato:
thats true 😄  forgot that :)

cpanato:
thanks again!

PR #342, opened 2018-03-13
update doc for testEmail

cpanato:
Server PR: https://github.com/mattermost/mattermost-server/pull/8453

update the doc to reflect the actual usage


v4/source/system.yaml
```diff
@@ -84,8 +91,17 @@
             Client := model.NewAPIv4Client("https://your-mattermost-url.com")
             Client.Login("email@domain.com", "Password1")
 
+            config := model.Config{
+                EmailSettings: model.EmailSettings{
```
saturninoabril: Just a minor comment: Indentation is not aligned. Might need to reduce by 2 spaces.

cpanato: done

PR #348, opened 2018-04-05
MM-9849 Added /config/environment route

hmhealey:
This route is needed to make fields in the system console read only when they've been set using environment variables. I'm not sure if the response schema makes sense here since the route is currently returning an object that looks like the config with `true` for settings set through environment variables with fields left out if they're not, so it looks like
```json
{
    "ServiceSettings": {
        "SiteURL": true
    },
    "EmailSettings": {
        "FeedbackName": true
    }
}
```

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-9849

### DIFF
--- a/v4/source/definitions.yaml
+++ b/v4/source/definitions.yaml
@@ -1264,6 +1264,498 @@ definitions:
           TurnSharedKey:
             type: string
 
+  EnvironmentConfig:
+    type: object
+    properties:
+      ServiceSettings:
+        type: object
+        properties:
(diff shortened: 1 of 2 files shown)


jwilander:
I think the response makes sense. The only suggestion I have regarding that is to make a new definition in the YAML for it, otherwise it's going to show string, etc. for the types of the response object

It would be nice if this could somehow be incorporated into the regular GET /config endpoint but I can't think of a good way to do that

PR #351, opened 2018-04-11
Remove invalid required paramter from oauth app PUT.

grundleborg:
Fixes the build of master.

Jenkins is still confused though, so thinks this PR doesn't build (it does, and it fixes the error).


jwilander:
Merging because this clearly doesn't have anything to do with the build failure

PR #356, opened 2018-04-16
MM-9990: Ephemeral Post Documentation

DSchalla:
This implements the documentation for change MM-9990 from https://github.com/mattermost/mattermost-server/pull/8611. Since the request body expects a structure that didn't exist as a schema I was a bit uncertain how to implement it, so if you'd prefer to create a dedicated schema for it or another proposal I am more than happy to change it.

### DIFF
--- a/v4/source/posts.yaml
+++ b/v4/source/posts.yaml
@@ -47,6 +47,47 @@
         '403':
           $ref: '#/responses/Forbidden'
 
+  /posts/ephemeral:
+    post:
+      tags:
+        - posts
+      summary: Create a ephemeral post
+      description: |
+        Create a new ephemeral post in a channel.
+        ##### Permissions
+        Must have `create_post_ephemeral` permission (currently only given to system adin)
+      parameters:
+        - in: body
+          name: ephemeral_post
+          description: Ephemeral Post object to send
+          required: true
+          schema:
+            type: object
+            required:
+              - user_id
+              - post
+            properties:
+              user_id:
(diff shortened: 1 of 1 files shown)


v4/source/posts.yaml
```diff
@@ -47,6 +47,47 @@
...
 
+  /posts/ephemeral:
+    post:
+      tags:
+        - posts
+      summary: Create a ephemeral post
+      description: |
+        Create a new ephemeral post in a channel.
+        ##### Permissions
+        Must have `create_post_ephemeral` permission (currently only given to system adin)
```
jwilander: Typo `adin` -> `admin`

v4/source/posts.yaml
```diff
@@ -47,6 +47,47 @@
...
+            properties:
+              user_id:
+                type: string
+                description: The target user id for the ephemeral post
+              post:
+                type: object
+                required:
+                  - channel_id
+                  - message
+                description: Post object to create
```
jwilander: This might work, but I'll think you'll have to specify properties for `channel_id` and `message`

DSchalla: Worked fine, but you are right, it should be added never the less!

DSchalla:
Thanks, fixed the typo and implemented the parameter descriptions. Build looks fine:
```
Building mattermost openapi yaml for v4
v4/html/static/mattermost-openapi-v4.yaml is valid
Complete
```

PR #360, opened 2018-04-25
MM-10182: Adds docs for update team scheme endpoint.

mkraft:
Docs for new API endpoint to update a team's scheme.

### DIFF
--- a/v4/source/teams.yaml
+++ b/v4/source/teams.yaml
@@ -969,7 +969,7 @@
         '400':
           $ref: '#/responses/BadRequest'
 
-'/teams/{team_id}/schemes':
+'/teams/{team_id}/scheme':
     put:
       tags:
         - teams
@@ -1026,7 +1026,7 @@
         - lang: 'curl'
           source: |
             curl -X PUT \
-              https://your-mattermost-url.com/api/v4/teams/4xp9fdt77pncbef59f4k1qe83o/schemes \
+              https://your-mattermost-url.com/api/v4/teams/4xp9fdt77pncbef59f4k1qe83o/scheme \
               -H 'Authorization: Bearer frn8fu5rtpyc5m4xy6q3oj4yur' \
               -H 'Content-Type: application/json' \
               -d '{"scheme_id": "qjda3stwafbgpqjaxej3k76sga"}'
--- a/v4/source/teams.yaml
+++ b/v4/source/teams.yaml
@@ -968,3 +968,65 @@
                 type: string
         '400':
           $ref: '#/responses/BadRequest'
+
+'/teams/{team_id}/schemes':
+    put:
+      tags:
+        - teams
+      summary: Set a team's scheme
+      description: |
+        Set a team's scheme, to wit sets the scheme_id value of a team record.
+        
+        ##### Permissions
+        Must have `manage_system` permission.
+
+        __Minimum server version__: 4.10
+      parameters:
+        - name: team_id
+          in: path
+          description: Team GUID
+          required: true
+        - in: body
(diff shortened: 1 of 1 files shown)


v4/source/teams.yaml
```diff
@@ -968,3 +968,58 @@
...
         '400':
           $ref: '#/responses/BadRequest'
+
+'/teams/{team_id}/schemes/{scheme_id}':
+    put:
+      tags:
+        - teams
+      summary: Set a team's scheme
+      description: |
+        Set a team's scheme.
```
jespino: Here I would be more explicit about what is really doing. Something like `Set the schema of the team with the id receive din the url as {team_id} to the schema with id received in the url as {schema_id}.`.

Without context is not so clear what is it doing.

v4/source/teams.yaml
```diff
@@ -968,3 +968,58 @@
                 type: string
         '400':
           $ref: '#/responses/BadRequest'
+
+'/teams/{team_id}/schemes/{scheme_id}':
+    put:
```
grundleborg: I don't think I've seen this pattern before in our APIs (but that might be just my ignorance) where we have URL parameters and an empty `PUT` to set some field on an object.

mkraft: I chose `PUT` because the request is idempotent and a zero length request body because it would be unnecessarily verbose (and isn't required by the spec). One objection to using a `PUT` [from the HTTP 1.1 spec](https://tools.ietf.org/html/rfc7231#section-4.3.4):

> A successful PUT of a given representation would suggest that a subsequent GET on that same target resource will result in an equivalent representation being sent in a 200 (OK) response.

...however it's just a suggestion and anyhow APIv4 does not adhering to that, for example:

* `/api/v4/teams/{{team_id}}/members/{{user_id}}/roles`
* `/api/v4/channels/{{channel_id}}/members/{{user_id}}/roles`
* `/api/v4/channels/{{channel_id}}/members/{{user_id}}/notify_props`

Did you have a specific argument for changing the HTTP verb or URL structure?

jespino: I think the weird pattern, at least for me, is passing the "new data" directly in the url, I would expect something like a PUT or a POST but in the url  `/teams/{team_id}/schemes` and with the body `{scheme_id: 'XXXXXXX'}`.

jwilander: Consistency is an important part of the API so I agree with @jespino, I'd prefer having it in the request body as that's what the rest of our endpoints do.

As for `PUT` vs `POST`, I'm with @mkraft on using `PUT` as that's what we use in the rest of the API for update semantics

mkraft: Well looks like it's 3 against 1 here so I've got some typing to do.

v4/source/teams.yaml
```diff
@@ -968,3 +968,65 @@
                 type: string
         '400':
           $ref: '#/responses/BadRequest'
+
+'/teams/{team_id}/schemes':
+    put:
```
grundleborg: Minor nitpick, but I think this should probably be `scheme` in the singular.

v4/source/teams.yaml
```diff
@@ -968,3 +968,65 @@
...
         '400':
           $ref: '#/responses/BadRequest'
+
+'/teams/{team_id}/scheme':
+    put:
+      tags:
+        - teams
+      summary: Set a team's scheme
+      description: |
+        Set a team's scheme, to wit sets the scheme_id value of a team record.
```
jwilander: `to wit sets` typo?

PR #361, opened 2018-04-25
MM-10183: Adds docs for update channel scheme endpoint.

mkraft:
Docs for new API endpoint to update a channel's scheme.


v4/source/channels.yaml
```diff
@@ -1330,3 +1330,56 @@
...
+      description: |
+        Set a channel's scheme.
+        
+        ##### Permissions
+        Must have `manage_system` permission.
+
+        __Minimum server version__: 4.10
+      parameters:
+        - name: channel_id
+          in: query
```
jespino: This must be path.

v4/source/channels.yaml
```diff
@@ -1330,3 +1330,56 @@
...
+          description: Channel GUID
+          required: true
+          type: string
+        - name: scheme_id
+          in: query
```
jespino: this must be path.

v4/source/channels.yaml
```diff
@@ -1330,3 +1330,56 @@
...
             // GetChannelUnread
             channelUnread, resp := Client.GetChannelUnread(<CHANNELID>, <USERID>)
+
+'/channels/{channel_id}/schemes/{scheme_id}':
+    put:
+      tags:
+        - channels
+      summary: Set a channel's scheme
+      description: |
+        Set a channel's scheme.
```
jespino: I would like to have a more clear explanation here. Without context is difficult to understand that you are setting the schema_id as schema of the channel.

v4/source/channels.yaml
```diff
@@ -1330,3 +1330,64 @@
...
             // GetChannelUnread
             channelUnread, resp := Client.GetChannelUnread(<CHANNELID>, <USERID>)
+
+'/channels/{channel_id}/scheme':
+    put:
+      tags:
+        - channels
+      summary: Set a channel's scheme
+      description: |
+        Set a channel's scheme, to wit sets the scheme_id value of a channel record.
```
grundleborg: very Shakespearean ;)

jwilander: I'd prefer if we used more common language here. We have many second/third language english readers using these docs and I'm going to be honest I've never seen this phrase before

mkraft: I'm going to leave it because it's a perfectly good way to say "more specifically".

jwilander: Is that the goal of these docs, though? To me the goal is to be easy to read and understand so I think we should use more common language wherever possible to achieve that goal.

grundleborg: I think Joram has a point here. It's quite an obscure phrase (I've never encountered it outside of the theatre/English Literature classes at school), so I think it probably doesn't belong in a technical document that's going to be read by many folks for whom English is a second/third/etc language.

mkraft: I don't agree, sorry.

esethna: Sorry @mkraft, I agree with JW and GG here. Can we use a more common phrase? Our docs are not localized so we need to be as straight forward as possible.

mkraft:
@esethna @grundleborg @jwilander Changed wording.

jwilander:
@mkraft mind rebasing this and then it should be good to merge

cpanato:
spoke in the pre-release channel

PR #368, opened 2018-05-15
MM-10605 Add int64 format to timestamps

jwilander:
https://mattermost.atlassian.net/browse/MM-10605


jwilander:
@cpanato any idea why the build is failing?

PR #370, opened 2018-05-18
Add descriptions for timestamp fields

chclaus:
I've also added description fields to the v3 api, although it's already deprecated


jwilander:
Awesome!

PR #371, opened 2018-05-19
Remove extra_update_at from Channels call

jasonblais:
This is for v5.0: https://github.com/mattermost/mattermost-server/pull/8762

I know the database column was deprecated but not removed..so if we should still keep this in the API doc, feel free to close this PR.

### DIFF
--- a/v4/source/definitions.yaml
+++ b/v4/source/definitions.yaml
@@ -133,7 +133,7 @@ definitions:
       total_msg_count:
         type: integer
       extra_update_at:
-        description: Removed in Mattermost 5.0 release
+        description: Deprecated in Mattermost 5.0 release
         type: integer
         format: int64
       creator_id:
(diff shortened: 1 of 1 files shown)


lieut-data:
@jasonblais, discussion at https://pre-release.mattermost.com/core/pl/ck1jhx5xbir7upmf6gmtijjw9w. Looks like you've already hit it via https://mattermost.atlassian.net/browse/MM-10649. 👍

jasonblais:
Gotcha 👍

jasonblais:
Makes sense, updated

v4/source/definitions.yaml
```diff
@@ -132,6 +132,10 @@ definitions:
         type: integer
       total_msg_count:
         type: integer
+      extra_update_at:
+        description: Removed in Mattermost 5.0 release
```
lieut-data: To be pedantic, since we're keeping both the column and the mention, `Deprecated in Mattermost 5.0 release` seems a bit clearer.

PR #377, opened 2018-06-14
MM-10928: fix /system/ping docs

lieut-data:
This references the /definitions/StatusOK ref to get the usual:
```
    {
      "status": "string"
    }
```

Note that the actual API currently returns additional fields referencing
the `ClientRequirements` struct from the configuration, but this
scheduled to be moved to a different endpoint and we're not encouraging
use of same.

### DIFF
--- a/v4/source/system.yaml
+++ b/v4/source/system.yaml
@@ -12,9 +12,7 @@
         '200':
           description: Status of the system
           schema:
-            type: object
-            items:
-              type: string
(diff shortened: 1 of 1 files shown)


enahum:
Im ok with this

hmhealey:
I don't think we need a new endpoint for those extra fields. They're already available in the client config, even with my changes, so we just need to switch it to read them from there

PR #378, opened 2018-06-22
Remove v3 from the api reference source code

jespino:


saturninoabril:
I can't remember if we have decisions to remove v3 API reference. Should we just put this in archive for historical reference?

jespino:
The decision AFAIK was to remove it from the repository, but keep the generated version on the web.

jwilander:
I removed the v3 publishing from the jenkins jobs as well

PR #380, opened 2018-07-04
MM-1118: document inability to delete direct and group message channels

lieut-data:
See https://github.com/mattermost/mattermost-server/pull/9054.

### DIFF
--- a/v4/source/channels.yaml
+++ b/v4/source/channels.yaml
@@ -296,7 +296,7 @@
         - channels
       summary: Delete a channel
       description: |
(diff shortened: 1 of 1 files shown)


v4/source/channels.yaml
```diff
@@ -296,7 +296,7 @@
         - channels
       summary: Delete a channel
       description: |
-        Soft deletes a channel, by marking the channel as deleted in the database. Soft deleted channels will not be accessible in the user interface.
+        Soft deletes a channel, by marking the channel as deleted in the database. Soft deleted channels will not be accessible in the user interface. Direct and group message channels may not be deleted.
```
jasonblais: propose `cannot be deleted` to remove ambiguity with `may`

PR #384, opened 2018-07-25
[MM-11210] Add docs for GET '/users/{user_id}/channels/{channel_id}/posts/unread'

saturninoabril:
Add docs for `GET '/users/{user_id}/channels/{channel_id}/posts/unread'`

Server PR https://github.com/mattermost/mattermost-server/pull/9108 is merged into `scrolling-overhaul` branch.

This is for dev review, but should not be merged until the `scrolling-overhaul` branch is merged into master.

Update:  Follow up PR - https://github.com/mattermost/mattermost-server/pull/9229

### DIFF
--- a/v4/source/posts.yaml
+++ b/v4/source/posts.yaml
@@ -397,6 +397,33 @@
         '403':
           $ref: '#/responses/Forbidden'
 
+  '/channels/{channel_id}/posts/unread':
+    get:
+      tags:
+        - posts
+      summary: Get posts for a channel around last unread
+      description: |
+        Get posts in a channel around last unread post of a member.
+        ##### Permissions
+        Must have `read_channel` permission for the channel.
+      parameters:
+        - name: channel_id
+          in: path
+          description: The channel ID to get the posts for
+          required: true
+          type: string
+      responses:
+        '200':
+          description: Post list retrieval successful
+          schema:
+            $ref: "#/definitions/PostList"
+        '400':
+          $ref: '#/responses/BadRequest'
+        '401':
+          $ref: '#/responses/Unauthorized'
+        '403':
+          $ref: '#/responses/Forbidden'
+
   '/teams/{team_id}/posts/search':
     post:
       tags:
--- a/v4/source/posts.yaml
+++ b/v4/source/posts.yaml
@@ -397,20 +397,36 @@
         '403':
           $ref: '#/responses/Forbidden'
 
-  '/channels/{channel_id}/posts/unread':
+  '/users/{user_id}/channels/{channel_id}/posts/unread':
     get:
       tags:
         - posts
       summary: Get posts for a channel around last unread
       description: |
         Get posts in a channel around last unread post of a member.
+        __Minimum server version: 5.3__
         ##### Permissions
         Must have `read_channel` permission for the channel.
       parameters:
+        - name: user_id
+          in: path
+          description: User GUID
+          required: true
+          type: string
         - name: channel_id
           in: path
-          description: The channel ID to get the posts for
+          description: Channel GUID
           required: true
+          type: string
+        - name: limit_before
+          in: query
+          description: The number of posts before the last unread post
+          default: "60"
+          type: string
+        - name: limit_after
+          in: query
+          description: The number of posts after and including the last unread post
+          default: "60"
           type: string
       responses:
         '200':


v4/source/posts.yaml
```diff
@@ -397,6 +397,33 @@
         '403':
           $ref: '#/responses/Forbidden'
 
+  '/channels/{channel_id}/posts/unread':
```
jwilander: Can we include the user ID in the path? That way a bot or an admin could get the unread posts for a different user if they needed to

hmhealey: Would we want that? Since this is a route designed specifically for one use case in our apps, it's doesn't seem as likely that other people will be using it. 0/5 though

v4/source/posts.yaml
```diff
@@ -397,6 +397,33 @@
...
+    get:
+      tags:
+        - posts
+      summary: Get posts for a channel around last unread
+      description: |
```
hmhealey: Can you add a note that this route will only be available in version 5.3 and later?

v4/source/posts.yaml
```diff
@@ -397,6 +397,33 @@
...
+        Get posts in a channel around last unread post of a member.
+        ##### Permissions
+        Must have `read_channel` permission for the channel.
+      parameters:
+        - name: channel_id
```
hmhealey: Is there a parameter to tell how many posts before/after to load?

saturninoabril: None at the moment.  It's using the default hard-coded value of 60 posts before (including the last unread) and 60 posts after.  I initially thought it's not necessary since we're not doing page request for this API.  However, I realized it makes sense to have.  Should I add something similar to "per_page", maybe like "limit"?

v4/source/posts.yaml
jwilander: My counter would be "why not?" I don't see a downside to adding the user ID to the path (if you want it for you you can use "me" instead of your user ID) and it gives more power to admins using the API. Also that would be consistent with other very similar endpoints we already have https://github.com/mattermost/mattermost-api-reference/blob/master/v4/source/channels.yaml#L1340

v4/source/posts.yaml
hmhealey: I think limit works better than per_page here. We could also make it limit_before and limit_after if we  wanted

v4/source/posts.yaml
hmhealey: I wasn't aware of the existing endpoint of that style. I think it makes sense even if I can't think of endpoints for it at this very moment

v4/source/posts.yaml
saturninoabril: Submitted PR to mattermost-server adding query of limit_before and limit_after.  Also updated this PR.

jwilander: Is there any reason why we might want to page through this endpoint? Or does that not make sense for the purpose of this endpoint?

v4/source/posts.yaml
```diff
@@ -397,6 +397,49 @@
         '403':
           $ref: '#/responses/Forbidden'
 
+  '/users/{user_id}/channels/{channel_id}/posts/unread':
+    get:
+      tags:
+        - posts
+      summary: Get posts for a channel around last unread
+      description: |
+        Get posts in a channel around last unread post of a member.
```
jwilander: Can we try to make this description more descriptive? To me it's not exactly clear what the purpose of this endpoint is

v4/source/posts.yaml
saturninoabril: For our webapp and mobile, maybe we're good at using the default values of 60 at the moment.  Those query parameters will give extra flexibility to other integrations.  0/5 with that though.

v4/source/posts.yaml
saturninoabril: Not sure if this gets any better.  I'm thinking something like:
```yaml
summary: Get posts for a channel around last unread
description: |
  This endpoint determines the last unread post of a member to the channel and gets a number of posts before and after the said post. Use the query parameters `limit_before` and `limit_after` to modify the number of posts to be returned by this endpoint. 
  __Minimum server version: 5.3__
  ##### Permissions
  Must have `read_channel` permission for the channel.
```

Let me know if that sounds good.  Would appreciate suggestion into it.  Thanks

hmhealey: `the said post` should just be `said post` or even `that post`. Also, I think you can leave out the explanation of the query parameters since they're explained below

v4/source/posts.yaml
```diff
@@ -397,6 +397,49 @@
...
+        __Minimum server version: 5.3__
```
hmhealey: Also, this needs to be 5.4 now

cpanato:
@saturninoabril this is still valid?

saturninoabril:
Yes, but I think it's better to close it now and then will just reopen next time.

PR #387, opened 2018-08-23
add channeltimezone api

cpanato:
#### Other Prs:
webapp: https://github.com/mattermost/mattermost-webapp/pull/1607
redux: https://github.com/mattermost/mattermost-redux/pull/612
server: https://github.com/mattermost/mattermost-server/pull/9286

also fix one issue

### DIFF
--- a/v4/source/channels.yaml
+++ b/v4/source/channels.yaml
@@ -196,6 +196,9 @@
       summary: Get timezones in a channel
       description: |
         Get a list of timezones for the users who are in this channel.
+
+        __Minimum server version__: 5.6
+
         ##### Permissions
         Must have the `read_channel` permission.
       parameters:
@@ -208,7 +211,9 @@
         '200':
           description: Timezone retrieval successful
           schema:
-            $ref: '#/definitions/Timezone'
+            type: array
+            items:
+              type: string
         '400':
           $ref: '#/responses/BadRequest'
         '401':
--- a/v4/source/channels.yaml
+++ b/v4/source/channels.yaml
@@ -189,6 +189,42 @@
             // GetPublicChannelsByIdsForTeam
             channels, resp := Client.GetPublicChannelsByIdsForTeam(<TEAMID>, channelIds)
 
+  /channels/{channel_id}/timezones:
+    get:
+      tags:
+        - channels
+      summary: Get timezones in a channel
+      description: |
+        Get a list of timezones for the users who are in this channel.
+        ##### Permissions
+        Must have the `read_channel` permission.
+      parameters:
+        - name: channel_id
+          in: path
+          description: Channel GUID
+          required: true
+          type: string
+      responses:
+        '200':
+          description: Timezone retrieval successful
+          schema:
+            $ref: '#/definitions/Timezone'
+        '400':
+          $ref: '#/responses/BadRequest'
+        '401':
+          $ref: '#/responses/Unauthorized'
+        '403':
+          $ref: '#/responses/Forbidden'
+      x-code-samples:
+        - lang: 'Go'
+          source: |
+            import "github.com/mattermost/mattermost-server/model"
+            Client := model.NewAPIv4Client("https://your-mattermost-url.com")
+            Client.Login("email@domain.com", "Password1")
+
+            // GetChannelStats
+            stats, resp := Client.GetChannelTimezones(<CHANNELID>)
+
   '/channels/{channel_id}':
     get:
       tags:
@@ -1401,7 +1437,7 @@
       summary: Set a channel's scheme
       description: |
         Set a channel's scheme, more specifically sets the scheme_id value of a channel record.
-        
+
         ##### Permissions
         Must have `manage_system` permission.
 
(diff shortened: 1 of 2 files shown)


v4/source/channels.yaml
```diff
@@ -189,6 +189,42 @@
             // GetPublicChannelsByIdsForTeam
             channels, resp := Client.GetPublicChannelsByIdsForTeam(<TEAMID>, channelIds)
 
+  /channels/{channel_id}/timezones:
+    get:
+      tags:
+        - channels
+      summary: Get users timezones for a channel
```
jwilander: Maybe update this to `Get timezones in a channel`

v4/source/channels.yaml
```diff
@@ -189,6 +189,42 @@
...
+      description: |
+        Get users timezones for a channel.
```
jwilander: And this to `Get a list of timezones for the users who are in this channel.`

v4/source/channels.yaml
```diff
@@ -189,6 +189,42 @@
...
+        Must have the `read_channel` permission.
+      parameters:
+        - name: channel_id
+          in: path
+          description: Channel GUID
+          required: true
+          type: string
+      responses:
+        '200':
+          description: Channel timezones retrieval successful
```
jwilander: And `Timezone retrieval successful`

v4/source/channels.yaml
```diff
@@ -189,6 +189,42 @@
...
+          schema:
+            $ref: '#/definitions/Timezone'
```
jwilander: Is this the right return? It just returns a single timezone?

cpanato: it return all timezones in that channel, can be 1 or more

v4/source/channels.yaml
cpanato: done

v4/source/channels.yaml
cpanato: done

v4/source/channels.yaml
cpanato: done

cpanato:
@jwilander PTAL

v4/source/channels.yaml
```diff
@@ -189,6 +189,42 @@
...
+        - channels
+      summary: Get timezones in a channel
+      description: |
+        Get a list of timezones for the users who are in this channel.
+        ##### Permissions
+        Must have the `read_channel` permission.
+      parameters:
+        - name: channel_id
+          in: path
+          description: Channel GUID
```
hmhealey: Huh, I didn't know that we referred to these as GUIDs in the API reference

v4/source/definitions.yaml
```diff
@@ -172,6 +174,14 @@ definitions:
         type: integer
         format: int64
 
+  ChannelTimezone:
```
hmhealey: This type isn't actually used

v4/source/channels.yaml
```diff
@@ -189,6 +189,42 @@
...
+          required: true
+          type: string
+      responses:
+        '200':
+          description: Timezone retrieval successful
+          schema:
+            $ref: '#/definitions/Timezone'
```
hmhealey: This should be an array instead of a single object. There's examples elsewhere in this file of how to do that

v4/source/channels.yaml
```diff
@@ -189,6 +189,42 @@
             // GetPublicChannelsByIdsForTeam
             channels, resp := Client.GetPublicChannelsByIdsForTeam(<TEAMID>, channelIds)
 
+  /channels/{channel_id}/timezones:
+    get:
+      tags:
+        - channels
+      summary: Get timezones in a channel
+      description: |
+        Get a list of timezones for the users who are in this channel.
```
hmhealey: In addition to the other things I asked about before, this should note that it requires server 5.6 or higher

v4/source/channels.yaml
cpanato: done

v4/source/definitions.yaml
cpanato: removed

v4/source/channels.yaml
cpanato: done

cpanato:
@jwilander @hmhealey PTAL

v4/source/definitions.yaml
```diff
@@ -2075,6 +2077,19 @@ definitions:
         type: string
         description: Set to "true" to enable mentions for first name. Defaults to "true" if a first name is set, "false" otherwise.
 
+  Timezone:
+    type: object
+    properties:
+      useAutomaticTimezone:
+        type: string
+        description: Set to "true" to use the browser/system timezone, "false"to set manually. Defaults to "true".
```
hmhealey:
```suggestion
        description: Set to "true" to use the browser/system timezone, "false" to set manually. Defaults to "true".
```

cpanato: done

cpanato:
@jwilander we will merge this only when release 5.6?

jwilander:
Nope, we have the note that the min server version is 5.6 so feel free to merge now

PR #389, opened 2018-09-08
Add auth_data and auth_service fields to POST /users API call

jasonblais:
It would be good for someone to verify my changes are technically accurate.

Credit to @xcompass for bringing this to our attention 👍

### DIFF
--- a/v4/source/users.yaml
+++ b/v4/source/users.yaml
@@ -4,7 +4,7 @@
         - users
       summary: Create a user
       description: |
-        Create a new user on the system.
+        Create a new user on the system. Password is required for email login. For other authentication types such as LDAP or SAML, auth_data and auth_service fields are required.
         ##### Permissions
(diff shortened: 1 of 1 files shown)


v4/source/users.yaml
```diff
@@ -39,7 +38,14 @@
                 type: string
               nickname:
                 type: string
+              auth_data:
+                description: Service-specific authentication data, such as email address.
+                type: string
+              auth_service:
+                description: The authentication service such as "email", "gitlab", or "ldap".
```
jwilander: Pretty sure this can also be `saml`

grundleborg: Looks like the currently supported options are: `email`, `gitlab`, `ldap`, `saml`, `office365`, `google`, or `` (i.e. empty)

jasonblais:
Thanks George, updated

PR #390, opened 2018-09-12
[MM-11593] Add a note to channel member delete endpoint

saturninoabril:
Add a note to channel member delete endpoint, like:
> Since server version 5.3, channel member can only be deleted either from public or private channel.

Jira ticket: [MM-11593](https://mattermost.atlassian.net/browse/MM-11593)


esethna:
Tweaked working slightly.

jwilander:
Made a minor change to be consistent with how we reference server versions in other places

PR #395, opened 2018-09-28
Adding the set default profile image endpoint

jespino:


v4/source/users.yaml
```diff
@@ -607,6 +609,65 @@
           $ref: '#/responses/NotFound'
         '501':
           $ref: '#/responses/NotImplemented'
+    delete:
+      tags:
+        - users
+      summary: Reset user's profile image to default image
+      description: |
+        Reset to default a user's profile image based on user_id string parameter.
```
saturninoabril: With the `delete` endpoint, word reset comes only second to me.  I'm thinking if it could be worded like, 
```
summary: Delete user's profile image
description: |
  Delete user's profile image and reset to default image based on user_id string parameter.
```

jespino: Fixed :)

grundleborg:
@saturninoabril @jespino is there anything holding back this PR from getting merged now?

saturninoabril:
Oops sorry, I didn't notice update on this.

PR #397, opened 2018-10-01
MM-12234: limit parameter on user search/autocomplete endpoint

lieut-data:
Document the new `limit` parameter added to the user search and and autocomplete endpoints: https://github.com/mattermost/mattermost-server/pull/9499.

### DIFF
--- a/v4/source/users.yaml
+++ b/v4/source/users.yaml
@@ -256,6 +256,10 @@
               without_team:
                 type: boolean
                 description: Set this to `true` if you would like to search for users that are not on a team. This option takes precendence over `team_id`, `in_channel_id`, and `not_in_channel_id`.
+              limit:
+                description: The maximum number of users to return in the results
+                type: integer
+                default: 100
       responses:
         '200':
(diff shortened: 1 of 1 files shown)


v4/source/users.yaml
```diff
@@ -256,6 +256,10 @@
               without_team:
                 type: boolean
                 description: Set this to `true` if you would like to search for users that are not on a team. This option takes precendence over `team_id`, `in_channel_id`, and `not_in_channel_id`.
+              limit:
+                description: The maximum number of users to return in the results
+                type: integer
+                default: 100
```
enahum: the redux client has this set to 25, do we want to change that or the default?

lieut-data: 0/5 here, but this is documenting the server defaults, not the Redux client defaults. I had picked `100` to preserve the semantics of the API as it exists today, but I anticipated clients would be converging to `25`.

lieut-data:
@hmhealey: done!

PR #401, opened 2018-10-04
Add documentation for the delete brand image action

jlevesy:
Hi folks.

In that PR I add documentation related to https://github.com/mattermost/mattermost-server/pull/9552.

Cheers !

### DIFF
--- a/v4/source/brand.yaml
+++ b/v4/source/brand.yaml
@@ -78,3 +78,35 @@
             }
 
             ok, resp := Client.UploadBrandImage(data.Bytes())
+
+    delete:
+      tags:
+        - brand
+      summary: Delete current brand image
+      description: |
+        Deletes the previously uploaded brand image. Returns 404 if no brand image has been uploaded.
+        ##### Permissions
+        Must have `manage_system` permission.
+        __Minimum server version: 5.6__
+      responses:
+        '200':
+          description: Brand image succesfully deleted
+          schema:
+            $ref: '#/definitions/StatusOK'
+        '401':
+          $ref: '#/responses/Unauthorized'
+        '403':
+          $ref: '#/responses/Forbidden'
+        '404':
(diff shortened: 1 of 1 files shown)


v4/source/brand.yaml
```diff
@@ -78,3 +78,30 @@
...
+          $ref: '#/responses/NotFound'
+      x-code-samples:
+        - lang: 'Go'
+          source: |
+            import "github.com/mattermost/mattermost-server/model"
+
+            Client := model.NewAPIv4Client("https://your-mattermost-url.com")
+            Client.Login("email@domain.com", "Password1")
+
+            // GetBrandImage
```
grundleborg: This should say `Delete Brand Image` I think.

v4/source/brand.yaml
```diff
@@ -78,3 +78,30 @@
...
+            img, err := Client.DeleteBrandImage()
```
grundleborg: This doesn't return the image, as the API endpoint only returns an "OK" response if it succeeds, not the image.

v4/source/brand.yaml
```diff
@@ -78,3 +78,30 @@
...
+        ##### Permissions
+        Must have `manage_system` permission.
+      responses:
+        '200':
+          description: Brand image succesfully deleted
+          schema:
+            $ref: '#/definitions/StatusOK'
+        '404':
+          description: No brand image uploaded
+          $ref: '#/responses/NotFound'
```
grundleborg: The server code appears to be able to return a `401` and `403` in some cases too.

grundleborg:
Hi @jlevesy . Thanks very much for submitting this PR. Just wondering how you are getting on with the review comments I left here? If there's anything unclear or that I can assist with, please do let me know - here or on pre-release (my username there is `@george`).

PR #402, opened 2018-10-11
Added API docs for GetUserTermsOfService API

harshilsharma:
Added API docs for GetUserTermsOfService API

### DIFF
--- a/v4/source/definitions.yaml
+++ b/v4/source/definitions.yaml
@@ -2221,7 +2221,7 @@ definitions:
         type: string
         description: The unique identifier of the terms of service the action was performed on.
       create_at:
-        description: The time in milliseconds the user performed action on this terms of service.
+        description: The time in milliseconds that this action was performed.
         type: integer
         format: int64
 
--- a/v4/source/definitions.yaml
+++ b/v4/source/definitions.yaml
@@ -2219,6 +2219,21 @@ definitions:
         type: string
         description: The text of terms of service. Supports Markdown.
 
+  UserTermsOfService:
+    type: object
+    properties:
+      user_id:
+        type: string
+        description: The unique identifier of the user who performed this terms of service action.
+      terms_of_service_id:
+        type: string
+        description: The unique identifier of the terms of service the action was performed on.
+      create_at:
+        description: The time in milliseconds the user performed action on this terms of service.
+        type: integer
+        format: int64
+
+
 externalDocs:
   description: Find out more about Mattermost
   url: 'https://about.mattermost.com'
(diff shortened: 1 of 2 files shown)


v4/source/users.yaml
```diff
@@ -1541,3 +1542,31 @@
           $ref: '#/responses/Unauthorized'
         '403':
           $ref: '#/responses/Forbidden'
+    get:
+      tags:
+      - users
+      - terms of service
+      summary: Fetches user's latest terms of service action if accepted.
```
grundleborg: I don't really understand this. Does it mean that it fetches the latest "accept" action of the terms of service by a user?

v4/source/users.yaml
```diff
@@ -1541,3 +1542,31 @@
...
+      description: |
+        Fetches user's latest terms of service action if accepted.
```
grundleborg: As above.

v4/source/definitions.yaml
```diff
@@ -2219,6 +2219,21 @@ definitions:
...
+    type: object
+    properties:
+      user_id:
+        type: string
+        description: The unique identifier of the user who performed this terms of service action.
+      terms_of_service_id:
+        type: string
+        description: The unique identifier of the terms of service the action was performed on.
+      create_at:
+        description: The time in milliseconds the user performed action on this terms of service.
```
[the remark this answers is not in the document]
harshilsharma: Agreed. Done.

v4/source/users.yaml
harshilsharma: We only store a user's latest action if it was for accepting the TOS, and delete it on rejection. So the API only returns the accepted action, if the latest action was for acceptance.

Perhaps the updated text makes sense?

v4/source/users.yaml
harshilsharma: Updated.

harshilsharma:
@hanzei I've created server PR with those docs - https://github.com/mattermost/mattermost-server/pull/9872

v4/source/users.yaml
```diff
@@ -1556,3 +1557,33 @@
           $ref: '#/responses/Unauthorized'
         '403':
           $ref: '#/responses/Forbidden'
+    get:
+      tags:
+      - users
+      - terms of service
+      summary: Fetches user's latest terms of service action if the latest action for for acceptance.
```
grundleborg: That feels a bit clearer to me now. Just a little typo to fix now (and below in the description). `for for acceptance` should probably be `was for acceptance`?

harshilsharma63: Yes, that's a typo. Fixed now.

PR #404, opened 2018-10-15
MM-12385: Fully document TeamMember object fields.

grundleborg:
https://mattermost.atlassian.net/browse/MM-12385

### DIFF
--- a/v4/source/definitions.yaml
+++ b/v4/source/definitions.yaml
@@ -275,10 +275,25 @@ definitions:
     type: object
     properties:
       team_id:
+        description: The ID of the team this member belongs to.
         type: string
       user_id:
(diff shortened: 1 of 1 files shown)


v4/source/definitions.yaml
```diff
@@ -275,10 +275,25 @@ definitions:
     type: object
     properties:
       team_id:
+        description: The ID of the team this member belongs to.
         type: string
       user_id:
+        description: The ID of the user this membership relates to.
```
jasonblais: Is it `membership` or `member`?

grundleborg: @jasonblais I changed it to `member`.

PR #405, opened 2018-10-17
MM-12505: Remove all things WebRTC

lieut-data:
Remove all things WebRTC. Complements https://github.com/mattermost/mattermost-webapp/pull/1907 and https://github.com/mattermost/mattermost-server/pull/9700.

### DIFF
--- a/v4/source/definitions.yaml
+++ b/v4/source/definitions.yaml
@@ -1344,25 +1344,6 @@ definitions:
         properties:
           MaxUsersForStatistics:
             type: integer
-      WebrtcSettings:
-        type: object
-        properties:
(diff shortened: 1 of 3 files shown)


hmhealey:
Do we maybe want to just mark those as deprecated instead of completely removing them from the reference? While I doubt this is the case for WebRTC in particular, it might be a good habit to not remove things completely since some people may be on older versions

crspeller:
@hmhealey I agree, but WebRTC is an exception. We are not following the normal deprecation process because it was never a production feature.

hmhealey:
Sounds good to me then. RIP WebRTC

PR #410, opened 2018-12-17
MM-13489 Add interactive dialog endpoints

jwilander:


v4/source/actions.yaml
```diff
@@ -0,0 +1,116 @@
...
+        __Minimum server version: 5.6__
+      parameters:
+        - name: body
+          in: body
+          description: Dialog submission data
+          required: true
+          schema:
+            type: object
+            required:
+              - url
```
saturninoabril: Is `url` required? I asked since SubmitDialogRequest says it's "omitempty".

v4/source/actions.yaml
```diff
@@ -0,0 +1,116 @@
...
+            required:
+              - trigger_id
+              - url
+              - dialog
+            properties:
+              trigger_id:
+                type: string
+                description: Trigger ID provided by other action
+              url:
+                type: object
```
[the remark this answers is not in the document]
jwilander: Should be a string, nice catch

v4/source/actions.yaml
jwilander: It's required, the omitempty is for when it's used in a different context

PR #415, opened 2019-01-02
Text updates for existing APIs

jasonblais:
These are based on report here https://github.com/mattermost/mattermost-server/issues/10023

### DIFF
--- a/v4/source/posts.yaml
+++ b/v4/source/posts.yaml
@@ -29,7 +29,7 @@
                 description: The post ID to comment on
               file_ids:
                 type: array
-                description: A list of file IDs to associate with the post
+                description: A list of file IDs to associate with the post. Note that uploads are limited to 5 files maximum. Please use additional posts for more files.
                 items:
                   type: string
               props:
--- a/v4/source/posts.yaml
+++ b/v4/source/posts.yaml
@@ -29,7 +29,7 @@
                 description: The post ID to comment on
               file_ids:
                 type: array
-                description: A list of file IDs to associate with the post. Note that uploads are limited to 5 files maximum. Please use additional posts for more files.
+                description: A list of file IDs to associate with the post. Note that posts are limited to 5 files maximum. Please use additional posts for more files.
                 items:
                   type: string
               props:
--- a/v4/source/files.yaml
+++ b/v4/source/files.yaml
@@ -44,7 +44,7 @@
           type: string
         - name: filename
           in: query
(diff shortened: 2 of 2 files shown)


v4/source/posts.yaml
```diff
@@ -29,7 +29,7 @@
                 description: The post ID to comment on
               file_ids:
                 type: array
-                description: A list of file IDs to associate with the post
+                description: A list of file IDs to associate with the post. Note that uploads are limited to 5 files maximum. Please use additional posts for more files.
```
lieut-data: I might be mistaken, but I don't think there is a limit to the number of files at the API level: the limit here is client-side only at the present time. That being said, there /is/ a limit within the API that relates to the total number of characters used by the filenames attached to the post, but of course it's variable.

Perhaps we just eliminate this part of the proposed update?
```suggestion
                description: A list of file IDs to associate with the post
```

DSchalla: It's kind of confusing right now. The following code checks for the max runes of the *file ids* submitted by the post: https://github.com/mattermost/mattermost-server/blob/master/model/post.go#L242

The ID has a length of 26 characters and is set by https://github.com/mattermost/mattermost-server/blob/master/model/file_info.go#L71 - So if we have 5 posts, we already got 130 characters. Some additional characters from JSON Encoding and we are close to 150 characters. So apparently there is indeed a limit enforced by the amount of files that can be attached, unless I am misinterpreting it.

Never the less, the error message "Invalid file ids" is not really intuitive, someone using the API won't know what went wrong.

jasonblais: In the original issue (https://github.com/mattermost/mattermost-server/issues/10023), it was reported that post creation fails when number of file IDs in list is more than 5.

Now that may be related to Daniel's note instead. Note that 6 files with an ID of 26 characters long is above 150 characters..

Regardless, I agree with Danielle that the existing error message is not obvious, which was separately updated here https://github.com/mattermost/mattermost-server/pull/10055. I'd be happy to update it to what is most appropriate for this case though.

lieut-data: Ah, interesting. So it does boil down to a maximum of five files per post.

My only suggestion, then, is to disassociate the term `upload`, since the files aren't being uploaded as part of this API: just being referenced.
```suggestion
                description: A list of file IDs to associate with the post. Note that posts are limited to 5 files maximum. Please use additional posts for more files.
```

jasonblais:
Suggestion looked good, thanks!

PR #421, opened 2019-02-11
Update posts.yaml

jasonblais:
It was reported that the props field for create post API is aa JSON object, not a string. If it's indeed a JSON object, wondering if this is incorrect elsewhere in the API docs

### DIFF
--- a/v4/source/posts.yaml
+++ b/v4/source/posts.yaml
@@ -34,7 +34,7 @@
                   type: string
               props:
                 description: A general JSON property bag to attach to the post
-                type: JSON object
+                type: object
       responses:
         '201':
           description: Post creation successful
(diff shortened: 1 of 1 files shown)


v4/source/posts.yaml
```diff
@@ -34,7 +34,7 @@
                   type: string
               props:
                 description: A general JSON property bag to attach to the post
-                type: string
+                type: JSON object
```
jasonblais: It seems we don't use a "JSON object" as a type elsewhere, so please do let me know if this is an incorrect term.

That said, the "string" type isn't really correct either.

crspeller: The correct type is "object" as above.

jasonblais: There we go, thanks Chris!

PR #432, opened 2019-03-22
Updated API docs to include TOS fields being included in certain user APIs

harshilsharma:
Updated API docs to include TOS fields being included in certain user APIs

### DIFF
--- a/v4/source/users.yaml
+++ b/v4/source/users.yaml
@@ -1654,7 +1654,7 @@
       - terms of service
       summary: Fetches user's latest terms of service action if the latest action was for acceptance.
       description: |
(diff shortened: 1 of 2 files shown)


v4/source/users.yaml
```diff
@@ -1654,6 +1654,7 @@
       - terms of service
       summary: Fetches user's latest terms of service action if the latest action was for acceptance.
       description: |
+        Will be depricated in v6.0
```
jasonblais:
```suggestion
        Will be deprecated in v6.0
```

hanzei:
@jasonblais Can this be merged?

Issue #429, opened 2019-02-28
'D' channel type is not documented


Psycojoker:
Hello,

Only 'O' and 'P' channel types are defined in the documentation here https://api.mattermost.com/#tag/channels

'D', which I though was for 'Deleted' (because their name was super weird), turned out to be 'Direct' and their name to be "{my-id}__{other-user-id}" which doesn't seems to be explained somewhere (I couldn't find this information)

Kind regards,

coltoneshaw:
@cpanato - I think this can be closed. The API doesn't return direct messages when hitting the channels endpoint, only O/P. 

Maybe this can be added to product docs?

PR #455, opened 2019-06-23
documentation: 'D' direct channel for "channel a update"

Ashid6:
description for "direct channel" should also come for "update a channel"

### DIFF
--- a/v4/source/channels.yaml
+++ b/v4/source/channels.yaml
@@ -353,7 +353,7 @@
                 description: Markdown-formatted text to display in the header of the channel
               type:
                 type: string
-                description: "'O' for a public channel, 'P' for a private channel"
+                description: "'O' for a public channel, 'P' for a private channel, 'D' for a direct channel"
       responses:
         '200':
           description: Channel update successful


ABresting:
Hi All, I am a GSoD'19 applicant. This is a minor fix but completes the channels.yml issue #429

Ashid6:
Hi All, I am a GSoD'19 applicant. This is a minor fix but completes the channels.yml issue #429

jwilander:
Hi @Ashid6, `D` is not actually a valid input for this API. To fix the linked ticket I think what we need to do is update the responses to any of the channel endpoints that return channels with a description on the `type` field explaining what the different values can be. Does that make sense?

Ashid6:
@jwilander yeah that makes sense. To update the responses I think we can change the definitions.yaml file  at https://github.com/mattermost/mattermost-api-reference/blob/master/v4/source/definitions.yaml#L141
here should we add "description:" with the types of the channel i.e. "O" and "P"

But the problem is #definitions/channel is used by many different types of requests, where not necessary  "O" and "P" types are used. either make inline change just every type of request or make new keys in definitions.yaml What do you think?

jwilander:
Sorry for the late response @Ashid6 

I'd suggest trying new definitions first. If they don't end up being very reusable then I think it would make sense to move to inline changes for each request

jasonblais:
Hey @Ashid6, let us know if you have any questions about the above? Happy to help :)

hanzei:
Closing this PR due to inactivity. Still, thanks for the contribution, @Ashid6! Let me know, if you still want to work on this.

PR #458, opened 2019-07-09
MM-15452 - Add docs for /bots/{bot_user_id}/icon

ali-farooq0:
### Summary
- Added docs for `GET`, `POST`, `DELETE` on `/bots/{bot_user_id}/icon` route (https://github.com/mattermost/mattermost-server/pull/11423)
- Fixed `bots` section as it wasn't showing up in the final api documentation page
- Added `make run` target to run `redoc` server for local testing
- Updated README to include instructions on how to test locally

### Docs preview
image

### README preview
image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,7 +54,7 @@ build-v4: .npminstall
 run:
 	@echo Starting redoc server
 
-	@node_modules/redoc-cli/index.js serve v4/html/static/mattermost-openapi-v4.yaml
+	npx redoc-cli serve v4/html/static/mattermost-openapi-v4.yaml
 
 clean:
 	@echo Cleaning
--- a/Makefile
+++ b/Makefile
@@ -54,7 +54,7 @@ build-v4: .npminstall
 run:
 	@echo Starting redoc server
 
-	redoc-cli serve v4/html/static/mattermost-openapi-v4.yaml
+	@node_modules/redoc-cli/index.js serve v4/html/static/mattermost-openapi-v4.yaml
 
 clean:
 	@echo Cleaning
--- a/Makefile
+++ b/Makefile
@@ -48,7 +48,13 @@ build-v4: .npminstall
 	@echo Getting dependencies using npm
 
 	npm install swagger-cli@1.0.0
+	npm install redoc-cli
 	touch $@
+
+run:
+	@echo Starting redoc server
+
+	redoc-cli serve v4/html/static/mattermost-openapi-v4.yaml
 
 clean:
 	@echo Cleaning
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ There is no strict style guide but please try to follow the example of the exist
 
 To build the full YAML, run `make build` and it will be output to `html/static/mattermost-openapi.yaml`. This will also check syntax using [swagger-cli](https://github.com/BigstickCarpet/swagger-cli).
 
-To test locally, run `make run` and navigate to `http://127.0.0.1:8080`.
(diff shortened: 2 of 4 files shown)


v4/source/bots.yaml
```diff
@@ -240,3 +238,148 @@
...
+
+  /bots/{bot_user_id}/icon:
+    get:
+      tags:
+        - bots
+      summary: Get bot's LHS icon
+      description: |
+        Get a bot's LHS icon image based on bot_user_id string parameter.
+        ##### Permissions
+        Must be logged in.
```
marianunez: Should we add here that they also need to have permission to view that bot user? 🤔

ali-farooq0: The only permission `get` requires is `view_members`. Which seems to be the default permission granted to all users. Didn't see other examples of it so i went with just the default description of `must be logged in`. Could change it if that's the new convention we want to use going forward.

lieut-data: I think it's fine to adopt the default description, but I suspect all of this needs to be updated to take into context guest accounts which explicitly don't have the usual permissions. cc @jespino

Makefile
```diff
@@ -48,8 +48,14 @@ build-v4: .npminstall
 	@echo Getting dependencies using npm
 
 	npm install swagger-cli@1.0.0
+	npm install redoc-cli
 	touch $@
 
+run:
```
lieut-data: 🎉

README.md
```diff
@@ -20,6 +20,8 @@ There is no strict style guide but please try to follow the example of the exist
 
 To build the full YAML, run `make build` and it will be output to `html/static/mattermost-openapi.yaml`. This will also check syntax using [swagger-cli](https://github.com/BigstickCarpet/swagger-cli).
 
+To test locally, run `make run` and navigate to `http://127.0.0.1:8080`.
```
lieut-data: Out of curiosity, does this also happen to enable real-time updates

ali-farooq0: Unfortunately not. I did update readme to include instructions on changes on the source file

Makefile
```diff
@@ -54,7 +54,7 @@ build-v4: .npminstall
 run:
 	@echo Starting redoc server
 
-	redoc-cli serve v4/html/static/mattermost-openapi-v4.yaml
+	@node_modules/redoc-cli/index.js serve v4/html/static/mattermost-openapi-v4.yaml
```
lieut-data: Is it possible to leverage npx vs. hard-coding the path?

v4/source/bots.yaml
jespino: Yes, exactly the `view_members` is granted to all users, but when guest accounts arrive, the guest users wont have that permission at all, so it only will be able to see users with direct relation (being in the same channel or being in a direct/group conversation).

Makefile
ali-farooq0: ah neat! done.

ali-farooq0:
@lieut-data all good to merge?

hanzei:
Hey all,

I'm a bit surprised this got merged before 5.14. I thought the convention is to merge new API endpoint PR's when the corresponding release is done.

lieut-data:
You're right, @hanzei -- my mistake. I'm inclined to leave it merged vs. reverting and re-merging later as such, especially given that we don't appear to adequately version api.mattermost.com anyway. Going forward, though, I'll try to re-adopt the existing pattern. Any objections?

hanzei:
No. I just wanted to make sure we are all on the same page when it comes to future changes.

PR #459, opened 2019-07-12
MM-11210 Add docs for GET '/users/{user_id}/channels/{channel_id}/posts/unread'

saturninoabril:
Add docs for GET '/users/{user_id}/channels/{channel_id}/posts/unread'

Server PR - https://github.com/mattermost/mattermost-server/pull/11486

### DIFF
--- a/v4/source/posts.yaml
+++ b/v4/source/posts.yaml
@@ -439,12 +439,12 @@
           type: string
         - name: limit_before
           in: query
-          description: Number of posts before the last unread posts
+          description: Number of posts before the last unread posts. Maximum is 200 posts if limit is set greater than that.
           default: "60"
           type: string
         - name: limit_after
           in: query
-          description: Number of posts after and including the last unread post
+          description: Number of posts after and including the last unread post. Maximum is 200 posts if limit is set greater than that.
           default: "60"
           type: string
       responses:
--- a/v4/source/posts.yaml
+++ b/v4/source/posts.yaml
@@ -426,6 +426,7 @@
         Get the oldest unread post in the channel for the given user as well as the posts around it.
         ##### Permissions
         Must be logged in as the user or have `edit_other_users` permission, and must have `read_channel` permission for the channel.
+        __Minimum server version__: 5.14
       parameters:
         - name: user_id
           in: path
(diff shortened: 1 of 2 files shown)


v4/source/posts.yaml
```diff
@@ -417,6 +417,48 @@
...
+          in: path
+          description: ID of the user
+          required: true
+          type: string
+        - name: channel_id
+          in: path
+          description: The channel ID to get the posts for
+          required: true
+          type: string
+        - name: limit_before
```
jwilander: Is there a maximum `limit_before` and `limit_after`?

saturninoabril: Yes, [max is 200](https://github.com/mattermost/mattermost-server/blob/master/web/params.go#L22).  I'll add that in description.

saturninoabril: Description added. Please take a look again, thanks

v4/source/posts.yaml
```diff
@@ -417,6 +417,48 @@
         '403':
           $ref: '#/responses/Forbidden'
 
+  '/users/{user_id}/channels/{channel_id}/posts/unread':
+    get:
+      tags:
+        - posts
+      summary: Get posts around last unread
+      description: |
+        Get posts around last unread post by a user in a channel. Use the query parameters `limit_before` and `limit_after` to modify the number of posts returned from this endpoint.
```
hmhealey: Just a suggestion to make the API's purpose a bit clearer. Also, I don't think we need to describe the parameters up here since they're explained below
```suggestion
        Get the oldest unread post in the channel for the given user as well as the posts around it.
```

saturninoabril: Thanks, applied the suggestion.

v4/source/posts.yaml
```diff
@@ -417,6 +417,48 @@
...
 
+  '/users/{user_id}/channels/{channel_id}/posts/unread':
+    get:
+      tags:
+        - posts
+      summary: Get posts around last unread
+      description: |
+        Get the oldest unread post in the channel for the given user as well as the posts around it.
+        ##### Permissions
+        Must be logged in as the user or have `edit_other_users` permission, and must have `read_channel` permission for the channel.
```
jwilander: Actually can we add the min server version?

saturninoabril: Thanks Joram, I forgot to push my changes last week.  Should be good now, please take a look again.

PR #460, opened 2019-07-21
Update introduction.yaml

Rajakavitha1:
Changed the title of a section so that the title is not redundant in the left navigation pane on : https://api.mattermost.com/#
redundant_title


lieut-data:
@jasonblais, agreed: I think `Overview` would make more sense as you propose. @Rajakavitha1, would you be open to help making that change?

Rajakavitha1:
@lieut-data and @jasonblais incorporated teh changes suggested.

PR #461, opened 2019-07-26
[MM-17194] Document the case for channel search where the user don't have the list_team_channels permission

mgdelacroix:
#### Ticket Link
https://mattermost.atlassian.net/browse/MM-17194

#### Related Pull Requets
- [Server PR](https://github.com/mattermost/mattermost-server/pull/11706)

### DIFF
--- a/v4/source/channels.yaml
+++ b/v4/source/channels.yaml
@@ -803,9 +803,9 @@
       description: |
         Search public channels on a team based on the search term provided in the request body.
         ##### Permissions
-        If the user has the `list_team_channels` permission, the search will be performed within all the team's public channels.
+        Must have the `list_team_channels` permission.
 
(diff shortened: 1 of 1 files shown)


v4/source/channels.yaml
```diff
@@ -803,7 +803,9 @@
       description: |
         Search public channels on a team based on the search term provided in the request body.
         ##### Permissions
-        Must have the `list_team_channels` permission.
+        If the user has the `list_team_channels` permission, the search will be performed within all the team's public channels.
+
+        Without the permission, the search will be performed within the channels that the user is a member of.
```
jwilander: Can you add a note about this only being the case in 5.16 and all previous versions you need the `list_team_channels` permission?

mgdelacroix: Added!

PR #464, opened 2019-08-01
[MM-17070] add set-unread to posts

Willyfrog:
Added docs for setting a channel as unread from a given post

Jira ticket: [MM-17070](https://mattermost.atlassian.net/browse/MM-17070)

### DIFF
--- a/v4/source/posts.yaml
+++ b/v4/source/posts.yaml
@@ -233,8 +233,10 @@
           required: true
           type: string
       responses:
-        '204':
+        '200':
           description: Post marked as unread successfully
+          schema:
+            $ref: "#/definitions/ChannelUnread"
         '400':
           $ref: '#/responses/BadRequest'
         '401':
--- a/v4/source/posts.yaml
+++ b/v4/source/posts.yaml
@@ -227,7 +227,7 @@
         Must have `read_channel` permission for the channel the post is in or if the channel is public, have the `read_public_channels` permission for the team.
         Must have `edit_user` permission if the user is not the one marking the post for himself.
 
-        __Minimum server version__: 5.14
+        __Minimum server version__: 5.16
       parameters:
         - name: user_id
           in: path
--- a/v4/source/posts.yaml
+++ b/v4/source/posts.yaml
@@ -225,8 +225,9 @@
         Mark a channel as being unread from a given post.
         ##### Permissions
         Must have `read_channel` permission for the channel the post is in or if the channel is public, have the `read_public_channels` permission for the team.
-        Currently `user_id` should be the same one logged in.
+        Must have `edit_user` permission if the user is not the one marking the post for himself.
 
+        __Minimum server version__: 5.14
       parameters:
         - name: user_id
           in: path
--- a/v4/source/posts.yaml
+++ b/v4/source/posts.yaml
@@ -216,17 +216,23 @@
         '403':
           $ref: '#/responses/Forbidden'
 
-  '/posts/{post_id}/set-unread':
+  '/user/{user_id}/posts/{post_id}/set_unread':
     post:
       tags:
         - posts
-      summary: Mark channel as not read.
+      summary: Mark as unread from the post.
       description: |
         Mark a channel as being unread from a given post.
         ##### Permissions
         Must have `read_channel` permission for the channel the post is in or if the channel is public, have the `read_public_channels` permission for the team.
+        Currently `user_id` should be the same one logged in.
 
       parameters:
+        - name: user_id
+          in: path
+          description: User GUID
+          required: true
+          type: string
         - name: post_id
           in: path
           description: Post GUID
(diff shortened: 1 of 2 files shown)


Willyfrog:
not sure if using that URL or somthing like `/post/<id>/channel/unread` which might be more consistent, but less readable.
Any thoughts?

hmhealey:
I'm not sure about that URL either. It's a bit redundant to include the channel ID, but I feel like it should have that as well.

There's a similar API to get the number of unread posts for a channel for a user ([link](https://api.mattermost.com/#tag/channels%2Fpaths%2F~1users~1%7Buser_id%7D~1channels~1%7Bchannel_id%7D~1unread%2Fget)), so we could use something similar to that. Maybe `POST /users/{user_id}/channels/{channel_id}/unread/{post_id}`?

Adding @jwilander to this PR since he might have some ideas here.

v4/source/posts.yaml
```diff
@@ -216,6 +216,36 @@
         '403':
           $ref: '#/responses/Forbidden'
 
+  '/posts/{post_id}/set-unread':
+    post:
+      tags:
+        - posts
+      summary: Mark channel as not read.
```
jwilander: What about documenting this as `Mark post as unread`? It's not /really/ just marking the post as unread since it will be marking anything newer than this post in the channel unread, but it makes it a little less weird since you're not providing a channel ID. We would just need to make sure that we are very clear about what's happening in the description below

v4/source/posts.yaml
```diff
@@ -216,6 +216,36 @@
         '403':
           $ref: '#/responses/Forbidden'
 
+  '/posts/{post_id}/set-unread':
```
jwilander:
```suggestion
  '/posts/{post_id}/set_unread':
```

Minor but should be underscores for consistency

jwilander: Actually, this is marking the channel/post unread for a specific user right? I assume you're just using the logged in user but we try to supply the user ID as part of API requests anyways so that admins/integrations with the right permissions could mark the posts as unread

So then it would be something like 
```suggestion
  '/users/{user_id}/posts/{post_id}/set_unread':
```

That make sense?

Willyfrog: I like the idea because of consistency.
One thing: do you want to implement controls for an admin to get to do that, or just theorising about possibilities?
I can't find any reason for an admin to do that, but you might have any.

hmhealey: We've gotten in the habit of including a user ID lately for new APIs in case someone comes up with a reason to do that later on

Willyfrog: yup, I got that, I was just curious if there was a use case, to make sure and implement it on the server :)
anyways it's already modified in both places to take the user id

jwilander: I think it's worth the extra effort to support admins specifying other user IDs. I could definitely see it used by an integration that works with unreads/notifications

v4/source/posts.yaml
```diff
@@ -216,6 +216,42 @@
...
+  '/user/{user_id}/posts/{post_id}/set_unread':
+    post:
+      tags:
+        - posts
+      summary: Mark as unread from the post.
+      description: |
+        Mark a channel as being unread from a given post.
+        ##### Permissions
+        Must have `read_channel` permission for the channel the post is in or if the channel is public, have the `read_public_channels` permission for the team.
+        Currently `user_id` should be the same one logged in.
```
hmhealey: We're going to have to support the user ID parameter since we're including it in the API route.

v4/source/posts.yaml
```diff
@@ -216,6 +216,43 @@
         '403':
           $ref: '#/responses/Forbidden'
 
+  '/user/{user_id}/posts/{post_id}/set_unread':
+    post:
+      tags:
+        - posts
+      summary: Mark as unread from the post.
```
jwilander:
```suggestion
      summary: Mark as unread from a post
```

v4/source/posts.yaml
```diff
@@ -216,6 +216,43 @@
...
+      description: |
+        Mark a channel as being unread from a given post.
+        ##### Permissions
+        Must have `read_channel` permission for the channel the post is in or if the channel is public, have the `read_public_channels` permission for the team.
+        Must have `edit_user` permission if the user is not the one marking the post for himself.
+
+        __Minimum server version__: 5.14
```
jwilander:
```suggestion
        __Minimum server version__: 5.16
```

This will be next release right?

v4/source/posts.yaml
```diff
@@ -216,6 +216,43 @@
...
+  '/user/{user_id}/posts/{post_id}/set_unread':
+    post:
+      tags:
+        - posts
+      summary: Mark as unread from the post.
+      description: |
+        Mark a channel as being unread from a given post.
+        ##### Permissions
+        Must have `read_channel` permission for the channel the post is in or if the channel is public, have the `read_public_channels` permission for the team.
+        Must have `edit_user` permission if the user is not the one marking the post for himself.
```
hmhealey: Shouldn't this be the `edit_other_users` permission?

jwilander:
Not stale, waiting for 5.16

PR #472, opened 2019-09-20
Fix failing go code in user API samples

streamer45:
#### Summary

Fixed a couple of problems with go code samples related to `User` API.


hanzei:
@jwilander Gentle reminder to review this. Should not take to long :wink:

PR #473, opened 2019-10-03
[MM-17754] add websocket event to docs

Willyfrog:
with the new mark as unread feature, there is a new websocket event: `post_unread`
I haven't marked a needed version as there seems to be no such feature for an item from a list


Willyfrog:
nevermind, I opened with the wrong branch

PR #478, opened 2019-10-28
[MM-13155] Finished documentation for the group endpoints

hahmadia:
Regarding https://mattermost.atlassian.net/browse/MM-13155

### DIFF
--- a/v4/source/groups.yaml
+++ b/v4/source/groups.yaml
@@ -6,9 +6,6 @@
       description: |
         Retrieve the list of groups.
 
-        ##### Permissions
-        Must have `manage_system` permission.
-
         __Minimum server version__: 5.11
       parameters:
         - name: page
@@ -20,6 +17,22 @@
           in: query
           description: The number of groups per page.
           default: "60"
+          type: string
+        - name: q
+          in: query
+          description: string to query groups by display name or name
+          type: string
+        - name: include_member_count
+          in: query
+          description: boolean which adds the `member_count` attribute to each group JSON object
+          type: boolean
+        - name: not_associated_to_team
+          in: query
+          description: string to exclude groups that are already linked to a given team via a `GroupTeams` table record
+          type: string
+        - name: not_associated_to_channel
+          in: query
+          description: string to exclude channels that are already linked to a given team via a `GroupChannels` table record
           type: string
       responses:
         '200':
--- a/v4/source/groups.yaml
+++ b/v4/source/groups.yaml
@@ -262,18 +262,6 @@
         '501':
           $ref: '#/responses/NotImplemented'
 
-
-
-
-
-
-
-
-
-
-
-
-
   '/groups/{group_id}/teams/{team_id}':
     get:
       tags:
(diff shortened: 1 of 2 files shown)


v4/source/groups.yaml
```diff
@@ -1,3 +1,563 @@
...
+        - name: page
+          in: query
+          description: The page to select.
+          default: "0"
+          type: string
+        - name: per_page
+          in: query
+          description: The number of groups per page.
+          default: "60"
+          type: string
```
mkraft: Other optional parameters include:
* `q` string to query groups by display name or name
* `include_member_count` boolean which adds the `member_count` attribute to each group JSON object
* `not_associated_to_team` string to exclude groups that are already linked to a given team via a `GroupTeam` table record.

v4/source/groups.yaml
```diff
@@ -1,3 +1,563 @@
...
+          required: true
+          type: string
+
+      responses:
+        '200':
+          description: Teams list retrieval successful
+          schema:
+            type: array
+            items:
+              $ref: '#/definitions/GroupSyncableTeams'
```
mkraft: I think you can just do this...
```suggestion
              $ref: '#/definitions/GroupSyncableTeam'
```
... and then delete `GroupSyncableTeams` definition because this is returned as a simple array.

v4/source/groups.yaml
```diff
@@ -1,3 +1,563 @@
...
+          required: true
+          type: string
+
+      responses:
+        '200':
+          description: Channel list retrieval successful
+          schema:
+            type: array
+            items:
+              $ref: '#/definitions/GroupSyncableChannels'
```
mkraft: Same thing here:
```suggestion
              $ref: '#/definitions/GroupSyncableChannel'
```

hahmadia:
Remove permissions /gropus

v4/source/groups.yaml
hahmadia: I thought the same thing until I was playing around with endpoints. For example, when I hit the `groups/:group_id/teams/:team_id` endpoint, I get the following:  
image

But when I hit `groups/:group_id/teams` endpoint, I get the following  
image

You can see a small discrepancy in the data returned (`team_display_name` and `team_type`) which I noticed which is why I created the two separate definitions. Should I omit the extra information and still revert it to a `GroupSyncableTeam`?

v4/source/groups.yaml
hahmadia: Same thing as above.
`/groups/:team_id/channels/:channel_id`  
image

`/groups/:team_id/channels`  
image  

Discrepancy in information received from each call hence the different definitions.

v4/source/groups.yaml
hahmadia: Thanks! Fixed.

v4/source/groups.yaml
mkraft: Oh I see. Yes I had forgotten that there were extra fields. You're totally right those fields should be documented. Nice work.

PR #481, opened 2019-11-01
document 404 when deleting plugins

lieut-data:
See https://github.com/mattermost/mattermost-server/pull/12983.


hanzei:
@lieut-data I think you need to merge `master` to fix CI

PR #482, opened 2019-11-06
MM-19337 API docs

valentijnnieman:
This documents the `search_archived` endpoint in [this PR](https://github.com/mattermost/mattermost-server/pull/12995).

### DIFF
--- a/v4/source/channels.yaml
+++ b/v4/source/channels.yaml
@@ -851,6 +851,62 @@
             // SearchChannels
             channels, resp := Client.SearchChannels(<TEAMID>, search)
 
+  '/teams/{team_id}/channels/search_archived':
+    post:
+      tags:
+        - channels
+      summary: Search archived channels
+      description: |
+        Search archived channels on a team based on the search term provided in the request body.
+        ##### Permissions
+        Must have the `list_team_channels` permission.
+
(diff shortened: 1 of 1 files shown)


v4/source/channels.yaml
```diff
@@ -851,6 +849,62 @@
...
+    post:
+      tags:
+        - channels
+      summary: Search archived channels
+      description: |
+        Search archived channels on a team based on the search term provided in the request body.
+        ##### Permissions
+        Must have the `list_team_channels` permission.
+
+        In server version 5.16 and later, a user without the `list_team_channels` permission will be able to use this endpoint, with the search results limited to the channels that the user is a member of.
```
hmhealey: This description should include `__Minimum server version__: 5.18` (assuming this makes it in before feature cutoff) since it's a new endpoint

hanzei:
I'm inclined to merge this PR now given that it's a new endpoint and its' clearly documented that this is for 5.20. Are you fine with this @crspeller @hmhealey?

hanzei:
cc @amyblais

amyblais:
@hanzei This is for v5.18. I think it's ok to merge if there is no risk of confusion.

hmhealey:
I'd be fine merging these doc changes even before the release as long as the feature isn't at risk of being cut. The docs specifically state the minimum version for the API endpoint anyway.

PR #487, opened 2019-11-12
Get channel members = channel memberships and roles

bradjcoughlin:
Updates docs to indicate that this endpoint returns a list of channels and associated roles for a user on a team, not a list of members (as the URL implies).

### DIFF
--- a/v4/source/channels.yaml
+++ b/v4/source/channels.yaml
@@ -1396,9 +1396,9 @@
     get:
       tags:
         - channels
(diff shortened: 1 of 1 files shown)


hmhealey:
The return type for this doesn't need to change, does it?

bradjcoughlin:
@hmhealey I don't believe so. It's still an array of objects (`ChannelMember` objects)

PR #489, opened 2019-11-25
MM-8607 - document serverbusy endpoints

wiggin77:
This pull-request provides endpoint documentation for PR https://github.com/mattermost/mattermost-server/pull/13212

### DIFF
--- a/v4/source/system.yaml
+++ b/v4/source/system.yaml
@@ -654,6 +654,9 @@
       summary: Set the server busy (high load) flag
       description: |
         Marks the server as currently having high load which disables non-critical services such as search, statuses and typing notifications.
+
+        __Minimum server version__: 5.20
+
         ##### Permissions
         Must have `manage_system` permission.
       parameters:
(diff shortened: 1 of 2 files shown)


v4/source/system.yaml
```diff
@@ -646,3 +646,90 @@
...
           $ref: '#/responses/Forbidden'
+
+  '/svrbusy':
+    post:
+      tags:
+        - system
+      summary: Set the server busy (high load) flag
+      description: |
+        Marks the server as currently having high load which disables non-critical services such as search, statuses and typing notifications.
+        ##### Permissions
```
jespino: I would add here, and in the other endpoints that the minimum server version is going to be 5.20

v4/source/system.yaml
```diff
@@ -646,3 +646,90 @@
...
+        '200':
+          description: Server busy flag set successfully
+          schema:
+            $ref: "#/definitions/StatusOK"
+        '400':
+          $ref: '#/responses/BadRequest'
+        '403':
+          $ref: '#/responses/Forbidden'
+      x-code-samples:
+        - lang: 'Go'
```
jespino: Can you add the `lang: 'curl'` option here?

v4/source/system.yaml
wiggin77: Minimum server version 5.20 added to all three endpoints via https://github.com/mattermost/mattermost-api-reference/commit/e26aa85745a0df9b21b3b642d3f7efeb1b05de95

v4/source/system.yaml
wiggin77: curl sample added to all three endpoints via https://github.com/mattermost/mattermost-api-reference/commit/e26aa85745a0df9b21b3b642d3f7efeb1b05de95 

curl samples tested as well.

hanzei:
I'm inclined to merge this PR now given that it's a new endpoint and its' clearly documented that this is for 5.20. Are you fine with this @reflog @jespino?

reflog:
Cool by me, maybe ping @amyblais to confirm?

amyblais:
@hanzei Is there a reason we want to merge this now instead of wait? It should be fine as long as there is no risk of confusion.

hanzei:
The reasons I see are:
1. We tent do forget merging these PRs.
2. The API endpoint is in the code now and devs might want to use it right now or know about it
3. To avoid merge conflicts.

PR #492, opened 2019-12-16
Doc for operations over non existent plugins

rvillablanca:
#### Summary

Add response codes 404 for operations over non existent plugins

- [x] Enabling plugin
- [x] Disabling plugin

For the case when deleting a non existent plugin it is already documented

#### Related issue

https://github.com/mattermost/mattermost-server/pull/11631

### DIFF
--- a/v4/source/plugins.yaml
+++ b/v4/source/plugins.yaml
@@ -228,6 +228,8 @@
           $ref: '#/responses/Unauthorized'
         '403':
           $ref: '#/responses/Forbidden'
+        '404':
+          $ref: '#/responses/NotFound'
         '501':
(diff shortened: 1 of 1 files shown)


rvillablanca:
@hanzei I have a couple of doubts.

1. I added to responde code list the status 404 but I'm not sure if it is necessary to add anything else.
2. In the plugin API was modified when getting the status of a non existent plugin, but I saw this is not exposed through the API so this doesn't require documentation. Let me know if I am wrong.

jasonblais:
@gabrieljackson Friendly reminder on review after the holidays :)

gabrieljackson:
Apologies for the delay on this. It looks good!

PR #503, opened 2020-03-17
Implemented SSR to improve load time / CPU usage

reflog:
#### Summary
Documentation is now prerendered and only JS for navigation/highlighting is used


reflog:
@saturninoabril - we can add it, but no real need. there's `package-lock.json` included and it's used with you do `npm i`

saturninoabril:
It didn't work for me.  `npm i` successfully fetched packages but didn't saved into node_modules. That's why `make build` then `make run` failed too.  I have a workaround for it but if current setup works for others, then it's all good for me.

reflog:
@saturninoabril - oh. well, in that case, I'm going to add package.json :)

PR #504, opened 2020-03-23
Add php samples for users documentation

gnello:
#### Summary
Add PHP samples for users endpoint documentation, based on the community-built [PHP Driver](https://github.com/gnello/php-mattermost-driver).

### DIFF
--- a/v4/source/users.yaml
+++ b/v4/source/users.yaml
@@ -165,7 +165,7 @@
             $driver = new Driver($container);
             $driver->authenticate();
 
-            $resp = $driver->getUserModel->createUser([
+            $resp = $driver->getUserModel()->createUser([
                 "username" => "username",
                 "email" => "email@domain.com",
                 "password" => "Password1"
@@ -637,7 +637,7 @@
             $channelID = "Ej3SKOHlWIKAblkUTK5Xvkj2cm";
             $channelID2 = "dWdfrUSdjJ7kyBvyBCgCav67Kz";
 
-            $resp = $driver->getUserModel->searchUsers([
+            $resp = $driver->getUserModel()->searchUsers([
                 "term" => "searchTerm",
                 "team_id" => $teamID,
                 "not_in_team_id" => $teamID2,
@@ -744,7 +744,7 @@
             $channelID = "Ej3SKOHlWIKAblkUTK5Xvkj2cm";
             $username = "testUsername";
 
-            $resp = $driver->getUserModel->autocompleteUsers([
+            $resp = $driver->getUserModel()->autocompleteUsers([
                 "team_id" => $teamID,
                 "channel_id" => $channelID,
                 "name" => $username,
@@ -804,7 +804,7 @@
             $driver = new Driver($container);
             $driver->authenticate();
 
-            $resp = $driver->getUserModel->getTotalCountOfUsersInTheSystem();
+            $resp = $driver->getUserModel()->getTotalCountOfUsersInTheSystem();
 
             if ($resp->getStatusCode() == 200) {
                 $stats = json_decode($resp->getBody())->total_users_count;
@@ -999,7 +999,7 @@
             $notifyProps = [];
             $notifyProps["comment"] = "somethingrandom";
 
-            $resp = $driver->getUserModel->updateUser($userID, [
+            $resp = $driver->getUserModel()->updateUser($userID, [
                 email => $email,
                 username => $username,
                 first_name => $firstName,
@@ -1077,7 +1077,7 @@
 
             $userID = "4xp9fdt77pncbef59f4k1qe83o";
 
-            $resp = $driver->getUserModel->deactivateUserAccount($userID);
+            $resp = $driver->getUserModel()->deactivateUserAccount($userID);
 
             if ($resp->getStatusCode() == 200) {
                $ok = json_decode($resp->getBody())->status;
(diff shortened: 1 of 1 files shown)


metanerd:
Thank you @gnello for your PR! Please ignore the pipeline problems.

v4/source/users.yaml
```diff
@@ -1093,6 +1369,26 @@
...
+                    "login_id" => "email@domain.com",
+                    "password" => "Password1",
+                  ]
+            ]);
+
+            $driver = new Driver($container);
+            $driver->authenticate();
+
+            $result = $driver->getUserModel()->setUserProfileImage("userid", [
+                "image" => fopen("profile_pic.png", 'rb'),
```
larkox: If I am not mistaken, here we are passing just the file identifier. Will this behave as expected?

v4/source/users.yaml
```diff
@@ -1405,6 +1797,24 @@
...
+                    "url" => "https://your-mattermost-url.com",
+                    "login_id" => "email@domain.com",
+                    "password" => "Password1",
+                  ]
+            ]);
+
+            $driver = new Driver($container);
+            $driver->authenticate();
+
+            $result = $driver->getUserModel()->generateMfaSecret("userid");
```
larkox: 0/5 Here I am missing something to identify the value we actually want. This happens on many examples, but this one felt more missing than others. I am not familiar with the `Gnello\Mattermost` code, but I would appreciate adding some line like:
```
if $resut->ok() {
   $mfaSecret = $result->content
}
```
Of course, previous lines adapted to what actually `Gnello\Mattermost` returns.

v4/source/users.yaml
```diff
@@ -2133,6 +2792,29 @@
...
+            ]);
+
+            $driver = new Driver($container);
+            $driver->authenticate();
+
+            $result = $driver->getUserModel()->switchLoginMethod([
+                "current_service" => "email",
+                "new_service" => "gitlab",
+                "email" => "test@domain.com",
+                //...
```
larkox: Unless there is a strong reason for leaving the rest of the values out of the example, I would put all the values on the Go example.

v4/source/users.yaml
```diff
@@ -2626,6 +3447,27 @@
...
+                    "password" => "Password1",
+                  ]
+            ]);
+
+            $driver = new Driver($container);
+            $driver->authenticate();
+
+            $result = $driver->getUserModel()->updateUserAuthenticationMethod("userid", [
+                "auth_service" => "gitlab",
+                //...
```
larkox: Unless there is a strong reason to not put all the fields, I would put all the fields in the Go example.

v4/source/users.yaml
```diff
@@ -276,6 +298,30 @@
...
+                    "url" => "https://your-mattermost-url.com",
+                    "login_id" => "email@domain.com",
+                    "password" => "Password1",
+                ]
+            ]);
+
+            $driver = new Driver($container);
+            $driver->authenticate();
+
+            $result = $driver->getUserModel()->getUsers([
```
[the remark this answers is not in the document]
gnello: Of corse yes, I'll try to make all the code look more like the go examples.
The "GetUsersNotInChannel" functionality can be achieved by passing (according to the api docs) the "not_in_team" params directly into the array, I'll show how in an separate example.

v4/source/users.yaml
```diff
@@ -502,6 +570,29 @@
...
+                    "url" => "https://your-mattermost-url.com",
+                    "login_id" => "email@domain.com",
+                    "password" => "Password1",
+                ]
+            ]);
+
+            $driver = new Driver($container);
+            $driver->authenticate();
+
+            $result = $driver->getUserModel()->searchUsers([
```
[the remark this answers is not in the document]
gnello: You're right, I agree.

v4/source/users.yaml
```diff
@@ -574,6 +665,29 @@
...
+            ]);
+
+            $driver = new Driver($container);
+            $driver->authenticate();
+
+            $result = $driver->getUserModel()->autocompleteUsers([
+                "team_id" => "teamid",
+                "channel_id" => "channelid",
+                "name" => "testUsername",
+                "limit" => 10,
```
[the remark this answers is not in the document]
gnello: No, any strong reason, I'll change it with 100.

v4/source/users.yaml
```diff
@@ -787,6 +962,24 @@
...
+                    "url" => "https://your-mattermost-url.com",
+                    "login_id" => "email@domain.com",
+                    "password" => "Password1",
+                  ]
+            ]);
+
+            $driver = new Driver($container);
+            $driver->authenticate();
+
+            $result = $driver->getUserModel()->deactivateUserAccount("userid");
```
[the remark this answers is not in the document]
gnello: Yes it does the same as DeleteUser, I know the naming suggest different things, it was inspired by the api docs summary.

v4/source/users.yaml
gnello: Yes, you're right but in the end it actually sends the binary file, as it mentioned [here](http://docs.guzzlephp.org/en/latest/quickstart.html#sending-form-files) in the http client documentation used by the php client. 
Anyway this "magic" is handled by the http client and I agree that it could be more clearer to provide a StreamInterface instead.

v4/source/users.yaml
gnello: Ok, I'll provide a result manipulation in all the examples. 
It will be a bit verbose due to the nature of the returned object ([here](https://github.com/gnello/php-mattermost-driver#check-results) an example) but I'll do my best.

larkox:
Thank you for the changes! LGTM :tada:

gnello:
I made one last check and corrected `$driver->getUserModel` into `$driver->getUserModel()`. Thanks.

hanzei:
@larkox Are you comfortable with the changes and they can get merged?

PR #518, opened 2020-04-30
MM-24692 - Update groups api to add new parameter 'since'

sbishel:
#### Summary
Adding parameter 'since' to getGroups api 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24692

#### Related PR
https://github.com/mattermost/mattermost-server/pull/14444


v4/source/groups.yaml
```diff
@@ -58,6 +58,15 @@
           required: true
           schema:
             type: string
+        - name: since
+          in: query
+          description: >
+            Only return groupss that have been modified since the given Unix
```
fmunshi, line 64:
```suggestion
            Only return groups that have been modified since the given Unix
```

sbishel: thanks

v4/source/groups.yaml
```diff
@@ -58,6 +58,15 @@
           required: true
           schema:
             type: string
+        - name: since
+          in: query
+          description: >
+            Only return groups that have been modified since the given Unix
+            timestamp (in milliseconds).
+
```
srkgupta, line 66: Can we update the documentation to indicate that this will cause the server to return even the deleted groups.

PR #520, opened 2020-05-01
[MM-24700] Add api docs for get groups by user ids

fmunshi:
#### Summary
- Adds docs for endpoint to get groups by user for a given list of user ids

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-24700


fmunshi:
Closing this as per discussion with UX Review we will no longer be needing `Groups` as a column on the team / channel members list - If it is decided that we in fact _do_ want it I can re-open this.

PR #527, opened 2020-05-29
MM-24425: Fix remaining references to non-v5 module

agnivade:
We fix any remaining code example using the mattermost-server/model
package to use the v5 module prefix.

Also, running npm install updated some package-lock.json entries. So fixing them.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-24425


agnivade:
@ogi-m - Not sure if there is a need to QA this. Let me know. Thanks.

PR #536, opened 2020-06-25
[MM-25749] Add dynamic autocomplete suggestions endpoint

iomodo:
#### Summary
This PR will add a description of the dynamic autocomplete suggestion's endpoint

#### Related PRs
server - https://github.com/mattermost/mattermost-server/pull/14915
mobile - https://github.com/mattermost/mattermost-mobile/pull/4484

### DIFF
--- a/v4/source/definitions.yaml
+++ b/v4/source/definitions.yaml
@@ -804,6 +804,18 @@ components:
         IconData:
           description: Base64 encoded svg image
           type: string
+    AutocompleteListItem:
+      type: object
+      properties:
+        Item: 
+          description: Predicted suggestion for a static list and dynamic arguments
+          type: string
+        Hint:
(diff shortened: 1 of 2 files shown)


v4/source/definitions.yaml
```diff
@@ -804,6 +804,18 @@ components:
...
           description: Base64 encoded svg image
           type: string
+    AutocompleteListItem:
+      type: object
+      properties:
+        Item: 
+          description: Predicted suggestion for a static list and dynamic arguments
+          type: string
+        Hint:
+          description: Hint about suggested input
```
levb, line 814: 0/5 when I was working on enabling Jira, I was a little puzzled as to what should be included in the hint; maybe an example and/or a style guide would be useful?

PR #541, opened 2020-07-02
MM-22417: Clarify behavior of /channels/{channel_id}/posts endpoint

agnivade:
It is known that the posts are actually returned from their update time
and not create time, and there can be gaps in the post order.

Although the gaps are not expected, but it is not known what will break
at this stage, so we just fix the documentation to clarify the behavior.


hmhealey:
Leaving to you to merge since I haven't seen Vercel before, and it scares me :p

agnivade:
@hmhealey - See the discussion in the DevOps team channel in core team :)

PR #569, opened 2020-08-24
Notices API

reflog:

### DIFF
--- a/v4/source/definitions.yaml
+++ b/v4/source/definitions.yaml
@@ -2896,6 +2896,15 @@ components:
     Notice:
       type: object
       properties:
+        id:
+          description: Notice ID
+          type: string
+        sysAdminOnly:
+          description: Does this notice apply only to sysadmins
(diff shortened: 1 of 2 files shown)


v4/source/definitions.yaml
```diff
@@ -2893,7 +2893,27 @@ components:
...
+          type: string
+        description:
+          description: "Notice content. Use {{Mattermost}} instead of plain text to support white-labeling. Text supports Markdown."
+          type: string
+        image:
+          description: URL of image to display
+          type: string
+        title:
+          description: "Notice title. Use {{Mattermost}} instead of plain text to support white-labeling. Text supports Markdown."
+          type: string
```
sudheerDev, line 2916: Would we need to add `sysAdminOnly` and `teamAdminOnly` flags here?

sudheerDev (changes requested):

sudheerDev (approved):

PR #572, opened 2020-09-01
[MM-27274] Update channel privacy docs

fmunshi:
### Summary
- Updates permission requirements for converting a channel privacy
- Related server https://github.com/mattermost/mattermost-server/pull/15370

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-27274

### DIFF
--- a/v4/source/channels.yaml
+++ b/v4/source/channels.yaml
@@ -724,7 +724,9 @@
 
         ##### Permissions
 
-        `manage_team` permission for the team of the channel.
+        `manage_team` permission for the channels team on version < 5.28.
+        `convert_public_channel_to_private` permission for the channel if updating privacy to 'P' on version >= 5.28.
+        `convert_private_channel_to_public` permission for the channel if updating privacy to 'O' on version >= 5.28.
       parameters:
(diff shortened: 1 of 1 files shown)


v4/source/channels.yaml
```diff
@@ -724,7 +724,9 @@
 
         ##### Permissions
 
-        `manage_team` permission for the team of the channel.
+        `manage_team` permission for the channels team on version < 5.28.
+        `convert_public_channel_to_private` permission for the channel if updating privacy to 'P' on version >= 5.28.
+        `convert_private_channel_to_public` permission for the channel if updating privacy to 'O' on version >= 5.28.
```
fmunshi, line 729: @reflog Is this okay? to have a different permission requirement on a different version and then document the changes as such? Or should I only show the latest permission requirements? Or does this break versioning? technically any role that had `manage_team` will now get the two new permissions so it shouldnt break any existing usage for anyone

reflog: I think it's good to document different versions. :+1:

reflog (approved):

PR #577, opened 2020-09-23
MM-27147 - Added entry for Cloud api

marianunez:
#### Summary

Added entry for initial Cloud api endpoints that include:

- GET `/cloud/products`
- POST `/cloud/payment/`
- POST /cloud/payment/confirm

#### Related PRs
Server: https://github.com/mattermost/mattermost-server/pull/15428

#### Ticket Link
[MM-27147](https://mattermost.atlassian.net/browse/MM-27147)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,6 +41,7 @@ build-v4: .npminstall
 	@cat $(V4_SRC)/reactions.yaml >> $(V4_YAML)
 	@cat $(V4_SRC)/actions.yaml >> $(V4_YAML)
 	@cat $(V4_SRC)/bots.yaml >> $(V4_YAML)
+	@cat $(V4_SRC)/cloud.yaml >> $(V4_YAML)
 	@cat $(V4_SRC)/definitions.yaml >> $(V4_YAML)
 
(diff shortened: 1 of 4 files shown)


jwilander:
I wonder if we should leave these out of the API docs and mark them in code with comments saying they're "internal and subject to change". Thoughts?

marianunez:
Agreed these do not make sense in the full api documentation given that they are not intended to be used outside of cloud but the build is not happy about that here: https://github.com/mattermost/mattermost-server/pull/15428

How do I get them to skip this validation?

ethervoid:
I had this conversation when I made the /login/cws endpoint, in the server channel and all of them prefer to have all the endpoints documented and public.

ethervoid (approved):
LGTM

jwilander:
I'm fine with having them documented and public. @marianunez maybe just add a line in each endpoint noting that they're meant for internal use and are subject to change

marianunez:
> I'm fine with having them documented and public. @marianunez maybe just add a line in each endpoint noting that they're meant for internal use and are subject to change

Note added.

jwilander (approved):

PR #583, opened 2020-10-22
[MM-29157] Invoices endpoints for cloud

ethervoid:
#### Summary

This PR the new cloud endpoints docs for invoices

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-29157

### DIFF
--- a/v4/source/cloud.yaml
+++ b/v4/source/cloud.yaml
@@ -241,7 +241,7 @@
         - cloud
       summary: Get cloud subscription invoices
       description: >
-        Retrieves the invoices for the subscriptio bound to this installation.
+        Retrieves the invoices for the subscription bound to this installation.
 
         ##### Permissions
 
--- a/v4/source/cloud.yaml
+++ b/v4/source/cloud.yaml
@@ -247,7 +247,7 @@
 
         Must have `manage_system` permission and be licensed for Cloud.
 
-        __Minimum server version__: 5.28
+        __Minimum server version__: 5.30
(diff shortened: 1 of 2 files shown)


v4/source/cloud.yaml
```diff
@@ -235,3 +235,64 @@
...
+        - cloud
+      summary: Get cloud subscription invoices
+      description: >
+        Retrieves the invoices for the subscriptio bound to this installation.
+
+        ##### Permissions
+
+        Must have `manage_system` permission and be licensed for Cloud.
+
+        __Minimum server version__: 5.28
```
marianunez, line 250: I think this should be 5.30? Not that it matters for us but technically is not in 5.28 because that was released already

marianunez (changes requested):

v4/source/cloud.yaml
```diff
@@ -235,3 +235,64 @@
           $ref: "#/components/responses/Forbidden"
         "501":
           $ref: "#/components/responses/NotImplemented"
+  /cloud/subscription/invoices:
+    get:
+      tags:
+        - cloud
+      summary: Get cloud subscription invoices
+      description: >
+        Retrieves the invoices for the subscriptio bound to this installation.
```
nickmisasi, line 244:
```suggestion
        Retrieves the invoices for the subscription bound to this installation.
```

nickmisasi (approved):
One nit, and I think Maria's right about 5.28 - otherwise, LGTM!

marianunez (approved):

PR #589, opened 2020-12-01
MM-31001 Add docs for remotecluster endpoints

wiggin77:
#### Summary
This PR adds documentation for three new endpoints:

1. api/v4/remotecluster/msg
2. api/v4/remotecluster/confirm_invite
3. api/v4/remotecluster/ping

The documentation is purposefully sparse because these endpoints can only be used by the new Remote Cluster Service.  This service allows clusters to communicate with each other.

Ideally this documentation wouldn't be needed as it pollutes the API reference with API's that cannot possibly be used by developers or users, however builds will fail without it due to linter rules.

@wiersgallak If you agree this does not belong in the API Reference, I could open a ticket to figure out how to allow exceptions to the linter.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-31001

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,6 +37,7 @@ build-v4: .npminstall
 	@cat $(V4_SRC)/roles.yaml >> $(V4_YAML)
 	@cat $(V4_SRC)/schemes.yaml >> $(V4_YAML)
 	@cat $(V4_SRC)/service_terms.yaml >> $(V4_YAML)
+	@cat $(V4_SRC)/remotecluster.yaml >> $(V4_YAML)
 	@cat $(V4_SRC)/opengraph.yaml >> $(V4_YAML)
 	@cat $(V4_SRC)/reactions.yaml >> $(V4_YAML)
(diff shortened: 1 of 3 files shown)


wiersgallak:
@wiggin77 Wondering if there are other endpoints that are documented that would cause a similar confusion point for customers?  If there are a few others, then we may want to take it on as a larger effort to clean up.

wiggin77:
> @wiggin77 Wondering if there are other endpoints that are documented that would cause a similar confusion point for customers? If there are a few others, then we may want to take it on as a larger effort to clean up.

@wiersgallak, I looked through the APIs and couldn't find any that I thought shouldn't be documented.  This will be the first time MM instances talk to each other via REST so it's not surprising these new APIs are the first instance of APIs that wouldn't be called by users/developers.

wiersgallak:
Thanks @wiggin77, I agree this should not be documented if a developer cannot program it.

wiggin77:
An exception has been added to MM govet so that these endpoints do not require documentation.

PR #601, opened 2021-01-21
MM-32013: users.yaml: update API definition to remove references to password

agnivade:
As per the current description, the API is only supposed to be used
to change authentication method. The password field was anyways a noop for that.
Removing that does not break anything. It can still be sent, but will remain as a noop.

https://mattermost.atlassian.net/browse/MM-32013

### DIFF
--- a/v4/source/definitions.yaml
+++ b/v4/source/definitions.yaml
@@ -578,6 +578,9 @@ components:
         auth_service:
           description: The authentication service such as "email", "gitlab", or "ldap"
           type: string
+      required:
+        - auth_data
+        - auth_service
     UserAutocomplete:
       type: object
       properties:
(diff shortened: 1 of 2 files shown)


v4/source/definitions.yaml
```diff
@@ -578,9 +578,6 @@ components:
         auth_service:
           description: The authentication service such as "email", "gitlab", or "ldap"
           type: string
```
srkgupta, line 580: Can we mark the auth_data and auth_service fields as required

agnivade: Sure, good one.

isacikgoz (approved):
LGTM 👍

srkgupta (approved):
Thanks @agnivade. Changes LGTM

Issue #586, opened 2020-11-19
PluginStatus.state has wrong property type


levee223:
The type of the PluginStatus.state property that has string enum values should be string, but it is number.

Fix it：
```diff
diff --git a/v4/source/definitions.yaml b/v4/source/definitions.yaml
index 3e404d1..202aa4b 100644
--- a/v4/source/definitions.yaml
+++ b/v4/source/definitions.yaml
@@ -2398,7 +2398,7 @@ components:
           type: string
           description: Path to the plugin on the server
         state:
-          type: number
+          type: string
           description: State of the plugin
           enum:
             - NotRunning
```

haardikdharma10:
@amyblais, can you confirm if this is a valid issue?

amyblais:
> can you confirm if this is a valid issue?

Maybe @hanzei would know?

hanzei:
@lameducks Good catch :+1: this find is valid.


@haardikdharma10 Would you be interested in submitting a fix?

haardikdharma10:
Yes, I would fix this @hanzei!

PR #611, opened 2021-02-16
[Fix - #586] : Change property type of PluginStatus.state from number to string.

haardikdharma10:
#### Summary

Fixes : #586

### DIFF
--- a/v4/source/definitions.yaml
+++ b/v4/source/definitions.yaml
@@ -2398,7 +2398,7 @@ components:
           type: string
           description: Path to the plugin on the server
         state:
-          type: number
+          type: string
           description: State of the plugin
           enum:
             - NotRunning


hanzei (approved):

lieut-data (changes requested):
Thanks for the recommendation, @haardikdharma10! Can you help me understand the problem here?

If I look at the definition of a `PluginStatus` in the model, the `state` is indeed a number that is described by various enumerated values:

https://github.com/mattermost/mattermost-server/blob/8e71ab1ecb4d05df5c13652bd3f0c8c4e8d9b0e6/model/plugin_status.go#L11-L29

Was there something you ran into that suggested otherwise? Or perhaps there is a different way to improve the documentation to reflect this?

lieut-data:
@haardikdharma10, let me know if this PR is still on your radar -- happy to continue the conversation to improve the developer experience here :)

haardikdharma10:
Sorry for the delay @lieut-data. After having a word with @hanzei once again, we decided it should be closed :)

PR #626, opened 2021-03-17
[MM-32044] Reset SAML auth data

maxerenberg:
#### Summary
This PR adds documentation for the API endpoints which resets SAML users' AuthData field to their email.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-32044

#### Server PR
https://github.com/mattermost/mattermost-server/pull/17161

### DIFF
--- a/v4/source/saml.yaml
+++ b/v4/source/saml.yaml
@@ -284,7 +284,7 @@
                   items:
                     type: string
                   default: null
-                  description: If set to an array, then users whose IDs are not in the array will be excluded.
+                  description: If set to a non-empty array, then users whose IDs are not in the array will be excluded.
       responses:
         "200":
           description: AuthData successfully reset
--- a/v4/source/saml.yaml
+++ b/v4/source/saml.yaml
@@ -252,3 +252,51 @@
           $ref: "#/components/responses/Forbidden"
         "501":
           $ref: "#/components/responses/NotImplemented"
+  /saml/resetid:
+    post:
+      tags:
+        - SAML
+      summary: Reset AuthData to Email
+      description: >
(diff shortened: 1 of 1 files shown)


v4/source/saml.yaml
```diff
@@ -252,3 +252,51 @@
...
+        - SAML
+      summary: Reset AuthData to Email
+      description: >
+        Reset the AuthData field of SAML users to their email. This is meant to be
+        used when the "id" attribute is set to an empty value ("") from a previously
+        non-empty value.
+
+        ##### Permissions
+
+        Must have `manage_system` permission.
```
sbishel, line 267: Add minimum server version. 

```suggestion
        Must have `manage_system` permission.
        
        __Minimum server version__: 5.35
```

v4/source/saml.yaml
```diff
@@ -252,3 +252,51 @@
...
+              type: object
+              properties:
+                include_deleted:
+                  type: boolean
+                  default: false
+                  description: Whether to include deleted users.
+                dry_run:
+                  type: boolean
+                  default: false
+                  description: If set to true, the number of users who would be affected is returned.
```
sbishel, line 281: Sorry, I missed this before and gave some bad advice. Dry-run should default to "true", but we don't want to default a boolean to true, so we should turn this flag into something like..."execute" and default it to false. That way they have to set a flag in order to manipulate the database.

sbishel (changes requested):

v4/source/saml.yaml
maxerenberg: As discussed offline, we will be keeping the API parameters the same as this would be more intuitive.

sbishel (approved):

PR #629, opened 2021-04-12
add dnd_end_time in requestBody of updateStatus

darkLord19:
Server PR: mattermost/mattermost-server#16067

### DIFF
--- a/v4/source/status.yaml
+++ b/v4/source/status.yaml
@@ -59,6 +59,9 @@
                 status:
                   type: string
                   description: User status, can be `online`, `away`, `offline` and `dnd`
+                dnd_end_time:
+                  type: integer
+                  description: Time on which dnd would be unset
         description: Status object that is to be updated
         required: true
       responses:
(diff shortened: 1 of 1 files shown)


v4/source/status.yaml
```diff
@@ -59,6 +59,9 @@
                 status:
                   type: string
                   description: User status, can be `online`, `away`, `offline` and `dnd`
+                dnd_end_time:
+                  type: integer
+                  description: Time on which dnd would be unset
```
agnivade, line 64: Time in epoch seconds at which a `dnd` status would be unset.

jespino (approved):
Nothing to add to the @agnivade suggestion :)

esethna:
@darkLord19 friendly ping on this PR so we can unblock the rest of them

darkLord19:
@esethna this is not a blocker. QA review needs to be done for server and webapp pr and if it goes fine then all 3 PRs can be merged

agnivade (approved):

PR #652, opened 2021-09-14
Clarify how to change system configuration

jpetazzo:
It took me a while to realize that the whole configuration should
be specified when doing a PUT (not just the fields that one wants
to change). Here is a small paragraph explaining that, hoping to
help others.

It looks like secret parameters (passwords, keys?) are automatically
replaced with strings of asterisks (`**********...`) and that these
get ignored when we PUT a new configuration, but I'm not sure.
If someone knows for sure, it would be fantastic to clarify as well.

Thanks! ♥

### DIFF
--- a/v4/source/system.yaml
+++ b/v4/source/system.yaml
@@ -404,6 +404,12 @@
         4.8, the `PluginSettings.EnableUploads` setting cannot be modified by
         this endpoint.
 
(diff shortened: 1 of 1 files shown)


cpanato:
thanks for your pr @jpetazzo

cpanato (approved):
lgtm

mkraft (approved):

PR #667, opened 2022-01-06
Clarified create user docs for authentication

coltoneshaw:
The API requires a personal access token when creating accounts using auth services such as LDAP.


cwarnermm (approved):
Thank you!

PR #668, opened 2022-01-06
[MM-36897] Update Fields for UpdateUser Endpoint

mylonsuren:
#### Summary

This PR updates the required fields to include `create_at`, `email`, and `username`. It also adds the `timezone` field.

#### Ticket Link

Fixes https://mattermost.atlassian.net/browse/MM-36897

### DIFF
--- a/v4/source/users.yaml
+++ b/v4/source/users.yaml
@@ -1153,9 +1153,14 @@
               type: object
               required:
                 - id
+                - create_at
+                - email
+                - username
               properties:
                 id:
                   type: string
+                create_at:
+                  type: integer
                 email:
                   type: string
                 username:
@@ -1170,6 +1175,8 @@
                   type: string
                 position:
                   type: string
+                timezone:
+                  $ref: "#/components/schemas/Timezone"
                 props:
                   type: object
                 notify_props:


cwarnermm (approved):
Thank you! Next steps include a technical review by Engineering.

v4/source/users.yaml
```diff
@@ -1153,9 +1153,14 @@
               type: object
               required:
                 - id
+                - create_at
```
mkraft, line 1156: You can update the createat field of a user? I'm wondering if that's actually a bug. Seems like something that shouldn't be updatable.

mylonsuren: Yeah it's a bit weird, the `create_at` field is required for the `UpdateUser` request, but you can't actually modify the field. So it's more of an issue with an unnecessary error being thrown when missing the `create_at` field rather than it being modifiable.

mkraft: I think that would be good to fix; sounds like we have a validation bug.

mylonsuren: Created an issue for this here [MM-41041](https://mattermost.atlassian.net/browse/MM-41041).

mkraft (approved):

mylonsuren:
Merging now that https://github.com/mattermost/mattermost-server/pull/19597 has been merged.

PR #675, opened 2022-02-14
Update delete channel description.

coltoneshaw:
#### Summary
Some confusion came up with a customer about the term soft delete. This means archived within the UI, so changing the lanquage to match what the customer would experience.

### DIFF
--- a/v4/source/channels.yaml
+++ b/v4/source/channels.yaml
@@ -629,7 +629,7 @@
       summary: Delete a channel
       description: >
         Archives a channel. This will set the `deleteAt` to the current timestamp in the database. Soft deleted channels may not be accessible in the user
-        interface. They can be viewed and unarchived in the system console > channel based on your license. Direct and group message channels cannot be deleted.
+        interface. They can be viewed and unarchived in the **System Console > User Management > Channels** based on your license. Direct and group message channels cannot be deleted.
 
 
         As of server version 5.28, optionally use the `permanent=true` query parameter to permanently delete the channel for compliance reasons. To use this feature `ServiceSettings.EnableAPIChannelDeletion` must be set to `true` in the server's configuration. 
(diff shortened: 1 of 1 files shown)


coltoneshaw:
@cwarnermm - Similar request. Needs review here as well. Thanks!

v4/source/channels.yaml
```diff
@@ -628,12 +628,12 @@
         - channels
       summary: Delete a channel
       description: >
-        Soft deletes a channel, by marking the channel as deleted in the
-        database. Soft deleted channels will not be accessible in the user
-        interface. Direct and group message channels cannot be deleted.
+        Archives a channel. This will set the `deleteAt` to the current timestamp in the database. Soft deleted channels may not be accessible in the user
+        interface. They can be viewed and unarchived in the system console > channel based on your license. Direct and group message channels cannot be deleted.
```
cwarnermm, line 632:
```suggestion
        interface. They can be viewed and unarchived in the **System Console > User Management > Channels** based on your license. Direct and group message channels cannot be deleted.
```

cwarnermm: I'm a bit unclear about the "based on your license" phrase in this description.

cwarnermm (commented):
Feedback provided inline and one clarifying question.

v4/source/channels.yaml
coltoneshaw: To my knowledge, not all license levels are able to do channel management within the system console. With all the license changes, I'm not sure where this line is drawn anymore.

cwarnermm: Great point! Channel management is available in [Mattermost Enterprise](https://docs.mattermost.com/configure/configuration-settings.html#channels). I've updated my recommendation to account for this.

PR #676, opened 2022-02-16
[MM-41679] Adding missing parameters

fromhro:
#### Summary
Adding missing GET parameters on /groups /teams /users endpoints

#### Ticket Link

[MM-41679](https://mattermost.atlassian.net/browse/MM-41679)

### DIFF
--- a/v4/source/groups.yaml
+++ b/v4/source/groups.yaml
@@ -69,6 +69,12 @@
             __Minimum server version__: 5.24
           schema:
             type: integer
+        - name: filter_allow_reference
+          in: query
+          description: Boolean which filters the group entries with the `allow_reference` attribute set.
+          schema:
+            type: boolean
+            default: false            
       responses:
         "200":
           description: Group list retrieval successful
@@ -866,6 +872,12 @@
             schema:
               type: boolean
               default: false
+          - name: paginate
+            in: query
+            description: Boolean do determine whether the pagination should be applied or not
+            schema:
+              type: boolean
+              default: false              
         responses:
           "200":
             description: Group list retrieval successful
--- a/v4/source/groups.yaml
+++ b/v4/source/groups.yaml
@@ -874,7 +874,7 @@
               default: false
           - name: paginate
             in: query
-            description: Boolean do determine whether the pagination should be applied or not
+            description: Boolean to determine whether the pagination should be applied or not
             schema:
               type: boolean
               default: false              
--- a/v4/source/users.yaml
+++ b/v4/source/users.yaml
@@ -1658,7 +1658,7 @@
             type: string
         - name: _
           in: query
-          description: unknown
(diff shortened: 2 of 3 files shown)


v4/source/groups.yaml
```diff
@@ -866,6 +872,12 @@
             schema:
               type: boolean
               default: false
+          - name: paginate
+            in: query
+            description: Boolean do determine whether the pagination should be applied or not
```
koox00, line 877: typo:

```suggestion
            description: Boolean to determine whether the pagination should be applied or not
```

koox00 (commented):
LG, just a typo!

v4/source/users.yaml
```diff
@@ -1656,6 +1656,12 @@
           required: true
           schema:
             type: string
+        - name: _
+          in: query
+          description: unknown
+          required: false
+          schema:
+            type: number
```
ashishbhate, line 1664: What's this section for?

fromhro: image
Goal is to describe in the Swagger this type of call that can actually be sent by MM client

ashishbhate: >Goal is to describe in the Swagger this type of call that can actually be sent by MM client 

Let's try to document the actual meaning of the parameter instead of just marking it as "unknown". 

Looking at the webapp code, https://github.com/mattermost/mattermost-webapp/blob/master/packages/mattermost-redux/src/client/client4.ts#L890

It seems the value of `_` is the `lastPictureUpdate`, but it isn't used on the server side. It's just a dummy parameter to get around any caching. 

I'm not sure if this needs to be documented in the API. 

@koox00 @Willyfrog what do you guys think?

Willyfrog: if it is a valid parameter, I'd document it, if only to indicate how to use it.

koox00: I would say that we probably shouldn't document it, since it has nothing to do with the API,
i.e it's not an accepted parameter, and might give the impression that the server actually consumes it and does something when it's passed, I'm 3/5.

@fromhro is that a blocker for your use case?

fromhro: @koox00 As there are some requests using this parameter on our end we would need to have it documented in the Swagger file. Otherwise our WAF policies won't allow the request using this parameter.

ashishbhate: Lets change:
```
description: unknown
```
to
```
description: Not used by the server. Clients can pass in the last picture update time of the user to potentially take advantage of caching,
```

fromhro: I did the modification

ashishbhate (approved):

ashishbhate (approved):
Thanks @fromhro, LGTM!

koox00 (approved):

PR #677, opened 2022-02-22
Updated channel endpoint parameters

coltoneshaw:
When investigating the system console > channels we noticed a few parameters that were not included in the documentation. These are those.

The query the system console uses is `http://<<siteURL>>/api/v4/channels?page=0&per_page=10&not_associated_to_group=&exclude_default_channels=false&include_total_count=true&include_deleted=true`

### DIFF
--- a/v4/source/channels.yaml
+++ b/v4/source/channels.yaml
@@ -10,7 +10,7 @@
       parameters:
       - name: not_associated_to_group
         in: query
-        description: Group GUID
+        description: A group id to exclude channels that are associated to that group via GroupChannel records. This can also be left blank with `not_associated_to_group=`.
         schema:
           type: string
       - name: page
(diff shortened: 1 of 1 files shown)


coltoneshaw:
@cwarnermm - Who would be the best reviewer for the API docs?

cwarnermm:
@coltoneshaw - Please feel free to add me as an editorial reviewer for all API updates.

coltoneshaw:
@cwarnermm - I don't seem to have permissions to update my PR after the fact. :/

v4/source/channels.yaml
```diff
@@ -10,7 +10,7 @@
       parameters:
       - name: not_associated_to_group
         in: query
-        description: Group GUID
+        description: A group id to exclude channels that are associated to that group via GroupChannel records. This can also be left blank with `not_associated_to_group=`.
```
cwarnermm, line 13:
```suggestion
        description: A group id to exclude channels that are associated with that group via GroupChannel records. This can also be left blank with `not_associated_to_group=`.
```

cwarnermm (approved):
One proposed suggestion. Approving for expediency.

v4/source/channels.yaml
```diff
@@ -29,6 +29,19 @@
         schema:
           type: boolean
           default: false
+      - name: include_deleted
+        in: query
+        description: Include channels that have been archived. This correlates to the DeleteAt flag being set in the database.
```
cwarnermm, line 34:
```suggestion
        description: Include channels that have been archived. This correlates to the `DeleteAt` flag being set in the database.
```

cwarnermm (approved):

PR #683, opened 2022-03-03
Replaced http URLs with https in various places

coltoneshaw:
#### Summary
After the http conversation I noticed a few urls are still http in this repo. I tried to get the ones that wouldn't cause an issue and updated them here. 

The `redoc-static.html` seems like it also has a ton of urls to update, but changing those didn't feed into the output `index.html`, so i'm not exactly sure where to make that change. It has a lot of URLS like `http://www.w3.org/2000/svg` that can be changed, and some in the scripts themselves.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,10 @@
 # Mattermost API Documentation
 
-This respository holds the API reference available at [https://api.mattermost.com](https://api.mattermost.com).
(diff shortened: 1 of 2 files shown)


emdecr (approved):
@coltoneshaw Nice catches! Looks good to me @cwarnermm 👍

cwarnermm (approved):
Thank you, @coltoneshaw!

PR #670, opened 2022-01-22
Add missing types to array definitions in channels and dataretention

gmerz:
#### Summary

Add missing item type to `team_ids` arrays in `SearchChannelsForRetentionPolicy` and `SearchAllChannels`. This provides a partial fix for #640

PR #662, opened 2021-11-07
Add operationId to /roles

unode:
#### Summary
Adds an `operationId` field to the API `/roles` call, arbitrarily called `GetAllRoles` by similarity to other API endpoints (e.g. `GetAllChannels`, `GetAllTeams`, ...).

PR #488, opened 2019-11-13
[MM-20015] Fix validation errors

iomodo:
This PR will fix validation errors thrown by https://editor.swagger.io
We are using [swagger-cli](https://github.com/BigstickCarpet/swagger-cli) to validate the yaml file, but it is not able to detect those errors.

Fixes: 
https://mattermost.atlassian.net/browse/MM-20015
https://github.com/mattermost/mattermost-api-reference/issues/468

PR #684, opened 2022-03-05
Use ibm-openapi-validator to validate api definition

gmerz:
#### Summary

This pull request 
+ introduces [Openapi validator](https://github.com/IBM/openapi-validator) for better validation of commits to the API and 
+ fixes issues found.

Currently builds are validated with swagger-cli. However, it seems that swagger-cli is missing many "easy to fix" things (See #488), which then create problems if you try to use the API with other tools (See #488).
The intention behind better validation is to remove the need for pull requests like #662 (missing operationID) or #670 (incomplete definitions of arrays).


v4/source/plugins.yaml
```diff
@@ -440,11 +443,13 @@
               type: object
               required:
                 - id
-                - version
```
agnivade, line 443: Wondering why are we removing this?

v4/source/webhooks.yaml
```diff
@@ -450,7 +450,7 @@
             schema:
               type: object
               required:
-                - id
+                - hook_id
```
agnivade, line 453: This is correct. It should be the opposite way. The json object field name is `id`. So we should update the path name to be `id` as well.

v4/source/webhooks.yaml
```diff
@@ -189,7 +189,7 @@
...
```
agnivade, line 192: Same as below

agnivade (commented):
Looks like a great improvement @gmerz ! Just some small comments to take care of.

v4/source/plugins.yaml
gmerz: Hi, because `version` is defined as "*required*", but there isn't even a "*version property*" defined in the following properties array. Only Id is defined. So, it can't be required if it isn't even a valid property. You know what I mean?
For the definition to be valid, either `version` should be removed from the required string, or added as a property to the following properties array.

I just wen't through the code again. While "version" would indeed be a valid field according to the struct definition [here](https://github.com/mattermost/mattermost-server/blob/9258a08b8da0fcc69b3b2e3a496d532023af11da/model/marketplace_plugin.go#L114), it is always [set to the empty string](https://github.com/mattermost/mattermost-server/blob/9258a08b8da0fcc69b3b2e3a496d532023af11da/api4/plugin.go#L156) before installing the plugin to always get the latest version (according to the comment there).

So, the version property could be passed to the endpoint, but it won't be used currently, as it seems. 
I have to admit, I don't know if parsing fails if it is not passed. But adding version to the property array to fix the API definition instead of removing it from required might still be the better option, even if it isn't used.

agnivade:
Thanks! I'll take another look.

cwarnermm:
@agnivade friendly reminder to review this PR. Thanks!

agnivade:
Sorry, I'll prioritize this.

agnivade (approved):
Apologies for the delay on this @gmerz. Looks good! I just pushed a small change to your PR. Hopefully that's okay with you. :)

PR #693, opened 2022-04-20
Clarify Websocket authentification methods

pheel:
#### Summary

Clarifies that authentication to Websocket can be performed using standard API authentication methods (and not just cookie or the challenge).

### DIFF
--- a/v4/source/introduction.yaml
+++ b/v4/source/introduction.yaml
@@ -250,7 +250,7 @@ tags:
       #### Authentication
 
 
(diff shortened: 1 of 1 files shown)


hmhealey (changes requested):
Thanks for the PR, and sorry that it got missed. I'll make sure we're actually following up on it from now on

hmhealey (approved):

hmhealey:
Thanks again!

PR #710, opened 2022-07-28
Post Reminders API Docs

ashishbhate:
#### Summary

- API documentation for post reminders
 
#### Ticket Link

- N/A

### DIFF
--- a/v4/source/posts.yaml
+++ b/v4/source/posts.yaml
@@ -873,3 +873,59 @@
           $ref: "#/components/responses/Unauthorized"
         "403":
           $ref: "#/components/responses/Forbidden"
+  "/users/{user_id}/posts/{post_id}/reminder":
+    post:
+      tags:
+        - posts
+      summary: Set a post reminder
+      description: >
+        Set a reminder for the user for the post.
+
+        ##### Permissions
+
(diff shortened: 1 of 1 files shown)


cwarnermm:
@ashishbhate - Is the post reminders feature intended to have a client-facing interface as well? If it does, then we may want to hold off publishing the API updates until this feature is available in-product. @agnivade - Thoughts?

agnivade:
I am 0/5, but the context here is that the server side PR is merged, but client side work is still going on. So technically speaking you can use the API using a CLI client. But obviously no client side interface is done yet.

No worries with holding off if you think so.

ashishbhate:
@cwarnermm Agniva summarised the context perfectly. I'll defer to whatever you think is best as well :)

cwarnermm:
Thank you both! I agree with @agnivade that since the server-side work is complete, and it's possible for customers to start using it, it makes sense to publish now rather than to wait!

Please note that we're currently experiencing issues with the API Reference site. Investigation is ongoing, but this content _may_ not be accessible until issues on the site have been addressed.

cwarnermm (approved):
Thank you, @ashishbhate!

agnivade (approved):

PR #715, opened 2022-08-18
[MM-44489] Cloud limits: enforcing files

vish9812:
#### Summary

Add `First-Inaccessible-File-Time` response-403-header for getFile APIs, which indicates if files are inaccessible due to the cloud plan's limit.

#### Ticket Link

JIRA - https://mattermost.atlassian.net/browse/MM-44489

Depends on this server PR - https://github.com/mattermost/mattermost-server/pull/20703

### DIFF
--- a/v4/source/files.yaml
+++ b/v4/source/files.yaml
@@ -145,7 +145,16 @@
         "401":
           $ref: "#/components/responses/Unauthorized"
         "403":
-          $ref: "#/components/responses/Forbidden"
+          description: Do not have appropriate permissions
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/AppError"
+          headers:
+            First-Inaccessible-File-Time:
+              schema:
+                type: integer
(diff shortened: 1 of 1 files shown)


neallred (approved):

v4/source/files.yaml
```diff
@@ -145,7 +145,16 @@
...
+          description: Do not have appropriate permissions
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/AppError"
+          headers:
+            First-Inaccessible-File-Time:
+              schema:
+                type: integer
+              description: This header is included with the value "1" if the post is past the cloud's plan limit.
```
ashishbhate, line 157: 0/5 - I think it would be more correct to say "file" instead of post here. Same applies for the other additions.

```suggestion
              description: This header is included with the value "1" if the file is past the cloud's plan limit.
```

ashishbhate (approved):

PR #718, opened 2022-08-24
changed user props field type

### LINKED ISSUE #643: User props is not fully described in API reference
Everywhere in the API reference User Props is described as `object`. For instance https://api.mattermost.com/#operation/CreateUser .
Regarding to https://www.json.org/json-en.html JSON object can have members of any JSON types.

But in official drivers the field `props` has type `map[string]string`
https://github.com/mattermost/mattermost-server/blob/d1eeafb29ff822cf6d9850a5307ab37c61fbc445/model/user.go#L85
https://github.com/mattermost/mattermost-redux/blob/3d1028034d7677adfda58e91b9a5dcaf1bc0ff99/src/types/users.ts#L42

Proposal:
In the API reference clearly specify that User Props can have members only of string type

1kjnv:
#### Summary

Changed User Props field type from "object" to map[string]string.

#### Ticket Link

Fixes #643

### DIFF
--- a/v4/source/bots.yaml
+++ b/v4/source/bots.yaml
@@ -764,7 +764,7 @@
                 position:
                   type: string
                 props:
-                  type: object
+                  $ref: "#/components/schemas/UserProps"
                 notify_props:
                   $ref: "#/components/schemas/UserNotifyProps"
         description: Data to be used in the user creation
--- a/v4/source/definitions.yaml
+++ b/v4/source/definitions.yaml
@@ -92,7 +92,7 @@ components:
         notify_props:
           $ref: "#/components/schemas/UserNotifyProps"
         props:
-          type: object
+          $ref: "#/components/schemas/UserProps"
         last_password_update:
           type: integer
         last_picture_update:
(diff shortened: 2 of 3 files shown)


plant99:
Hi @ikjnv , @cwarnermm apologies for the delay.

I was thinking about this PR offline, and realized that mattermost-server doesn't maintain a schema for this.

So, `props` like `notify_props` can take a much broader object than the two fields mentioned here. It'd take some time for me to collect all of them.

This PR looks very good to me, but would it mislead the users regarding extent of `user.props`?

cwarnermm:
Thanks, @plant99, for reviewing this PR. @laneycs - Can a member of the Suite Users Team review and advise on next steps for this PR? I'm specifically looking for guidance on whether the content update proposed in this PR is sufficient, or if there's better value to our users if we document the broader capabilities of these two objects. If it's the latter, I'll need some help from the team to document this information.

laneycs:
@cwarnermm Sure! Will add to our triage backlog to review with the team.

cwarnermm:
@laneycs - Does your team have an ETA on when this PR will be technically reviewed?

laneycs:
@cwarnermm Looks like our team still needs to triage this ticket. You can follow its progress here: https://mattermost.atlassian.net/browse/MM-46897. We triage incoming items in our backlog every Tuesday and try to make room for dedicated maintenance and tech debt work 1 day per week. Hoping we're able to triage and take action on it in the next couple of weeks. 

Heads up that some of our maintenance work has been pushed back as we have instructions from MLT to develop packaging changes for the 7.6 release, so we've had to make a bit of a pivot in our roadmap.

cwarnermm:
Thank you, @laneycs! Really appreciate the context and link to the Jira ticket driving this effort.

PR #728, opened 2023-03-13
last_post_at timestamp should be int64

### LINKED ISSUE #727: Number format in yaml specifies integer type for last_post_at but returns int64
This is the same issue that was mentioned in #345 but I guess this `last_post_at` field (under `Channel`) was added later on.

patryk-s:
#### Summary

The `last_post_at` field under `Channel` is a timestamp. Therefore, it should be formatted as `int64`.

EDIT: Also fixed `last_password_update` and `last_picture_update` as [suggested below](https://github.com/mattermost/mattermost-api-reference/pull/728#pullrequestreview-1437032048)

#### Ticket Link

Fixes #727

### DIFF
--- a/v4/source/definitions.yaml
+++ b/v4/source/definitions.yaml
@@ -94,9 +94,9 @@ components:
         props:
           type: object
         last_password_update:
-          type: integer
+          type: int64
         last_picture_update:
-          type: integer
+          type: int64
         failed_attempts:
           type: integer
         mfa_active:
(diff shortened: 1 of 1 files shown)


marianunez (approved):
Thanks @patryk-s! 

I've also noticed that we may need to include this for `last_password_update` and `last_picture_update` of `User` as defined here: https://github.com/mattermost/mattermost-server/blob/0cb3a406da7a339cc47bb72e32106b24e13c2a9a/server/public/model/user.go#L91

Would you be open to include in this PR as well?

PR #732, opened 2023-04-17
Absolute paths in API Spec

lieut-data:
#### Summary
Switch to prefixing `/api/v4` explicitly on all API paths, allowing the dynamically pulled in Playbooks API to anchor itself at `/plugins/playbooks/api/v0` without ambiguity. Even without the Playbooks, this anchor allows us to more easily distinguish a potential `/api/v5` in the future.

Before:

image
image

After:

image
image

#### Ticket Link
None, but see https://github.com/mattermost/mattermost-server/pull/22989.

### DIFF
--- a/v4/source/introduction.yaml
+++ b/v4/source/introduction.yaml
@@ -642,6 +642,6 @@ x-tagGroups:
       - exports
       - usage
 servers:
-  - url: http://your-mattermost-url.com/api/v4
-  - url: https://your-mattermost-url.com/api/v4
+  - url: http://your-mattermost-url.com
+  - url: https://your-mattermost-url.com
 paths:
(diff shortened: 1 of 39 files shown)


v4/source/introduction.yaml
```diff
@@ -642,6 +642,6 @@ x-tagGroups:
       - exports
       - usage
 servers:
-  - url: http://your-mattermost-url.com/api/v4
-  - url: https://your-mattermost-url.com/api/v4
+  - url: http://your-mattermost-url.com
+  - url: https://your-mattermost-url.com
```
lieut-data, line 646: This is the "bit that counts" relative to what we wanted to fix. Note that Playbooks had its own servers, but they got clobbered with the merge.

cwarnermm (approved):
This is a good path forward. Thanks, @lieut-data!

koox00 (approved):

